PR #16, opened 2026-07-01
Self-contained Spec of optimizer correctness

georgwiese:
*Builds on #19*

This PR adds a self-contained `Spec.lean`, specifying what it means for a circuit optimizer to be correct. The intention is that it should be easy to audit, minimal, and complete. The idea here is that in the end all files that humans need to review is:
- `Spec.lean`
- A zkVM-specific instantiation of `BusSemantics`
- Performance benchmarks

Everything else can be managed autonomously by AI.

### DIFF
--- a/Leanr/Spec.lean
+++ b/Leanr/Spec.lean
@@ -6,6 +6,8 @@ set_option autoImplicit false
 -- `p` is the field characteristic; a prime, so `ZMod p` is a field.
 variable {p : ℕ} [Fact p.Prime]
 
+--------- Expressions ---------
+
 /-- An arithmetic expression over variables (identified by `String`) and field constants. -/
 inductive Expression (p : ℕ) where
   | const (n : ZMod p)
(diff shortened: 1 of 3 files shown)


Leanr/Spec.lean
```diff
@@ -0,0 +1,131 @@
...
+/-- Evaluate a bus interaction under an assignment `env`, turning a symbolic bus
+    interaction into a bus interaction message. -/
+def BusInteraction.eval (bi : BusInteraction (Expression p)) (env : String → ZMod p) :
+    BusInteraction (ZMod p) :=
+  { busId := bi.busId,
+    multiplicity := bi.multiplicity.eval env,
+    payload := bi.payload.map (fun e => e.eval env) }
+
+/-- The bus semantics of the zkVM. -/
+structure BusSemantics (p : ℕ) where
```
georgwiese, line 44: #17 Adds bus semantics for OpenVM

georgwiese: Also, in the [`autoopt`](https://github.com/powdr-labs/leanr/tree/autoopt) branch, I [ended up relaxing this a little bit](https://github.com/powdr-labs/leanr/commit/805abe80132b1f33b0821c77833bff219145bda1#diff-f0d01e1a880b4ab71fbcefbdac127ba0acd1cc9f6fe8a9031d04e1c7044e93ab) to enable optimizations on the memory bus that require additional assumptions. I think it makes sense to open a separate PR for that though.

Leanr/Spec.lean
```diff
@@ -0,0 +1,131 @@
...
+/-- A constraint system representing a single zkVM chip. -/
+structure ConstraintSystem (p : ℕ) where
+  algebraicConstraints : List (Expression p)
+  busInteractions : List (BusInteraction (Expression p))
+
+/-- The side effects of a constraint system under a given environment and bus semantics.
+    The side effects are the tuples sent to the *stateful* buses.-/
+def ConstraintSystem.sideEffects (cs : ConstraintSystem p)
+    (busSemantics : BusSemantics p) (env : String → ZMod p) : BusState p :=
+  cs.busInteractions.filter (fun bi => busSemantics.isStateful bi.busId)
```
leonardoalt, line 84: 🟠 **HIGH: equivalentTo can ignore witness constraints and stateless bus obligations**  _(correctness)_

`sideEffects` drops all non-stateful bus interactions, and `implies` relates `self` and `other` through independently chosen satisfying environments whose only link is equivalence of those filtered side effects. For systems without stateful buses, this reduces equivalence to mutual satisfiability, so purely algebraic systems with disjoint solution sets can be deemed equivalent; stateless lookup or range-check obligations can likewise be deleted while still satisfying `equivalentTo`. This makes `optimizerMaintainsCorrectness` unsound because an optimizer can be certified while accepting witnesses that do not correspond to the original system.

**Suggested fix:** Strengthen `implies` so satisfying assignments correspond over the shared or public variables, for example by using a shared witness or requiring agreement between environments, and include stateless bus obligations in the compared observable effect.

— _reviewr multi-AI consensus_

Leanr/Spec.lean
```diff
@@ -0,0 +1,131 @@
...
+    let message := bi.eval env
+    message.multiplicity ≠ 0 → busSemantics.breaksInvariant message = false
+
+/-- Whether a constraint system implies another under a given bus semantics.
+    Informally, a constraint system `self` implies a system `other` if for every
+    satisfying assignment of `self`:
+    1. There exists a satisfying assignment of `other`
+    2. The side effects of `self` under the given environment and bus semantics
+       are equal to the side effects of `other` under the corresponding environment. -/
+def ConstraintSystem.implies (self other : ConstraintSystem p) (busSemantics : BusSemantics p) :
```
leonardoalt, line 109: 🟠 **HIGH: equivalentTo can ignore witness constraints and stateless bus obligations**  _(correctness)_

`sideEffects` drops all non-stateful bus interactions, and `implies` relates `self` and `other` through independently chosen satisfying environments whose only link is equivalence of those filtered side effects. For systems without stateful buses, this reduces equivalence to mutual satisfiability, so purely algebraic systems with disjoint solution sets can be deemed equivalent; stateless lookup or range-check obligations can likewise be deleted while still satisfying `equivalentTo`. This makes `optimizerMaintainsCorrectness` unsound because an optimizer can be certified while accepting witnesses that do not correspond to the original system.

**Suggested fix:** Strengthen `implies` so satisfying assignments correspond over the shared or public variables, for example by using a shared witness or requiring agreement between environments, and include stateless bus obligations in the compared observable effect.

— _reviewr multi-AI consensus_

Leanr/Spec.lean
```diff
@@ -0,0 +1,131 @@
...
+  | .add e1 e2 => e1.eval env + e2.eval env
+  | .mul e1 e2 => e1.eval env * e2.eval env
+
+--------- Bus Interactions ---------
+
+/-- A bus interaction. Typically, α is
+    - an expression (=> symbolic bus interaction), or
+    - a field element (=> bus interaction message). -/
+structure BusInteraction (α : Type) where
+  busId : Nat
```
leonardoalt, line 31: 🟡 **MEDIUM: Spec cannot represent symbolic bus IDs accepted by the codebase**  _(correctness)_

`BusInteraction` fixes `busId : Nat`, while the legacy representation treats bus IDs as expressions: JSON parsing reads `id` through `parseJsonExpr`, and solver logic substitutes and refines `busId` like other expression fields. Proofs over this spec therefore do not cover optimizers that rewrite, substitute, or depend on symbolic bus-id expressions.

**Suggested fix:** Either keep the legacy shape with an expression-typed bus ID and evaluate it with the payload, or add a proved normalization layer showing that all legacy bus IDs are constants before applying this spec.

— _reviewr multi-AI consensus_

Leanr/Spec.lean
```diff
@@ -0,0 +1,131 @@
...
+abbrev BusState (p : ℕ) := List (BusMessage p × ZMod p)
+
+/-- The net multiplicity with which `message` is sent in `state`. -/
+def multiplicitySum (message : BusMessage p) (state : BusState p) : ZMod p :=
+  match state with
+  | [] => 0
+  | (msg, mult) :: tl => (if msg = message then mult else 0) + multiplicitySum message tl
+
+/-- Two bus states are equal when every message is sent with the same net multiplicity. -/
+instance : HasEquiv (BusState p) :=
```
leonardoalt, line 70: 🔵 **LOW: HasEquiv instance leaks onto raw List type via abbrev reducibility**  _(maintainability)_

`BusState p` is a reducible abbrev for `List ((Nat, List (ZMod p)), ZMod p)`, so the `HasEquiv` instance is registered for the underlying `List` shape rather than a distinct semantic bus-state type. This makes the equivalence instance apply globally to any list of that shape, which can cause surprising instance resolution outside the intended domain.

**Suggested fix:** Make `BusState` a `def` or a one-constructor wrapper structure so the instance keys to a distinct head type.

— _reviewr multi-AI consensus_

Leanr/Spec.lean
```diff
@@ -0,0 +1,131 @@
+import Mathlib.Data.ZMod.Basic
+
+set_option autoImplicit false
+
+-- `p` is the field characteristic; a prime, so `ZMod p` is a field.
+variable {p : ℕ} [Fact p.Prime]
```
leonardoalt, line 6: ⚪ **NIT: Unused Fact p.Prime assumption**  _(maintainability)_

The comment says `p` is prime so `ZMod p` is a field, but the file does not use primality; the definitions only require the ring operations available for `ZMod p`. Keeping an unused standing assumption can mislead auditors into thinking field structure is part of the argument.

**Suggested fix:** Drop the unused `Fact p.Prime` assumption here, or clarify that it is only a convention for downstream instantiations.

— _reviewr multi-AI consensus_

leonardoalt (commented):
## 🤖 reviewr — multi-AI consensus review

Independent reviews by **claude, codex**, then blind cross-review over **2 rounds** (reached consensus). 4 agreed finding(s); 5 posted inline below.

The spec's equivalence relation is too weak to certify optimizer correctness: it permits independent satisfying witnesses and compares only filtered stateful side effects, so systems with different accepted assignments or missing stateless bus interactions can be considered equivalent. It also narrows legacy symbolic bus IDs to constants, leaving part of the existing optimizer surface outside the proof, with a few smaller maintainability issues in the Lean encoding.

**Verdict:** Request changes: the core equivalence relation is unsound for proving optimizer correctness.

<details><summary>Run log</summary>

```
Fetching PR #16 head into a throwaway worktree…
  linked .lake -> /home/leo/devel/leanr/.lake
Reviewing: Self-contained Spec of optimizer correctness (legacy-move...spec)
Inspecting code in: /tmp/reviewr-pr16-3fp6nxt7
Reviewers: claude, codex
Artifacts: /home/leo/devel/reviewr/.reviewr/runs/powdr-labs-leanr-pr16-20260702-214031

Round 0: independent review
  claude: 2 findings
  codex: 2 findings
Round 1: debate (4 shared findings)
  +1 new · contested=False · stable=False
Round 2: debate (5 shared findings)
  +0 new · contested=False · stable=True
Converged after round 2.
Composing final review
Wrote /home/leo/devel/reviewr/.reviewr/runs/powdr-labs-leanr-pr16-20260702-214031/REVIEW.md
```

</details>

Leanr/Spec.lean
georgwiese: It is true that the legacy one supported it and part of powdr does as well, but I don't think it's needed in practice and our [`SymbolicBusInteraction`](https://github.com/powdr-labs/powdr/blob/main/autoprecompiles/src/symbolic_machine.rs#L71-L75) also doesn't support expressions. So I think this is fine and can be generalized later if needed. Anything that's too strict should be fine.

Leanr/Spec.lean
georgwiese: sounds good to me

Leanr/Spec.lean
georgwiese: > For systems without stateful buses, this reduces equivalence to mutual satisfiability

Yes, this is a deliberate choice: We only care about the change it can make to the machine state. For example, if you take a sound circuit and add completely unrelated variables and constraints (no stateful bus interactions), they are considered equivalent under the current definition.

Leanr/Spec.lean
georgwiese: I thought AIs being caught in a loop is a thing of the past...

leonardoalt: I guess this was a reviewr bug, there were probably two findings on the same line.

PR #51, opened 2026-07-07
Task 2: derived columns in Leanr (represented, threaded, emitted, serialized)

leonardoalt:
## Summary

Represents powdr's **derived columns** (witgen hints) in Leanr, threads them through the pipeline,
**emits them from the re-encoding pass**, and serializes them in powdr's exact format — with all
correctness proofs closed (**no `sorry`, no `admit`, no new axiom**). `lake build` is fully green;
`openVmOptimizer_maintainsCorrectness` depends only on `[propext, Classical.choice, Quot.sound]`.

⚠️ **Touches the audited surface** (`Leanr/Spec.lean` `refines`/`impliesAdmissible`,
`Leanr/Optimizer.lean`). Per `CLAUDE.md` this is a correctness-spec change and warrants human review
(not a merge-without-review effectiveness PR). Effectiveness (variables/bus/constraints) is
**byte-identical** to the hint-free pipeline.

## Data model (mirrors powdr)

- `ComputationMethod p` = `constant (ZMod p)` | `quotientOrZero (num den)` (field-inverse, `0` on
  zero divisor) | `ifEqZero (cond) (thenM elseM)`; `eval` reproduces powdr's
  `evaluate_computation_method`.
- `DerivedVariable p = { isNew : Bool, name : String, computation : ComputationMethod p }`.
- `ConstraintSystem` gains `derivedColumns : List (DerivedVariable p) := []`.

## Spec strengthening (why it stays sound)

Derived columns are witgen hints, not verifier constraints — `satisfies`/`implies`/`withinDegree`
ignore them. The **completeness** direction of `refines` (`impliesAdmissible`) is strengthened to
*assume* `self.derivedConsistent env` and *deliver* a witness `env'` with
`other.derivedConsistent env'`. The assume/deliver symmetry keeps `refl` (env'=env re-delivers the
assumption) and `trans` (middle witness carries the next hypothesis) closing, so **all pass
composition is unchanged**. Soundness (`implies`) is untouched: an adversarial output assignment
can't be made hint-consistent and needn't be. The 4th clause of `optimizerMaintainsCorrectness` is
now `optimizerDerivedColumnsConsistent` — a projection of the strengthened `refines`: for every
admissible/satisfying/hint-consistent input there's an output-satisfying `env'` making every emitted
hint consistent (exactly what witgen needs following the hints on a real trace).

This is *not* env-universal entailment, which is **unprovable** for hints of eliminated variables
(free in the output). Consistency under the *constructed completeness witness* is the provable,
[...]

PR #56, opened 2026-07-07
Task 3: combine FFI + derived columns

leonardoalt:
## Task 3: combine the FFI and derived-columns efforts, run the benchmark

Combines the two open Leanr efforts and wires a witgen-correct Lean optimizer into powdr for a Lean-vs-Rust benchmark. Companion powdr PR: powdr-labs/powdr#3785. Builds on #53 (FFI) and #51 (derived columns).

### What's here
- **Merged** `claude/ffi-lean-optimizer` (#53) and `claude/derived-columns` (#51). The only conflict was that both branches independently added `Leanr/Implementation/JsonSerializer.lean`; reconciled into one serializer that (a) matches powdr's serde for the full `SymbolicMachine` and (b) emits `derived_columns` as the 3-tuple `[isNew, "name@id", <ComputationMethod>]` with an externally-tagged `ComputationMethod` (`{"Constant":n}` / `{"QuotientOrZero":[num,den]}` / `{"IfEqZero":[cond,then,else]}`). The fresh-variable renaming map is applied to derived-column names and to the expressions inside their computation methods too.
- **Witgen-safe FFI path.** The audited spec's `optimizerPreservesDerivedColumns` conjunct forces the optimizer to carry derived columns through **verbatim**, so a pass *cannot* emit a witgen hint without an audited-`Spec.lean` change (out of scope) — and even then the pass framework imposes no `derivedColumnsEntailed` obligation, so emission wouldn't be provably correct. `reencodePass` is the only pass that mints fresh witness columns (the re-encoding bits), which powdr witgen can only fill from such a hint. So `Leanr/Ffi.lean` now runs `optimizerWithBusFactsNoReencode` — the identical pipeline minus `reencodePass`, **still a fully verified refinement** (it reuses the same `VerifiedPassW` combinators, each bundling its own `PassCorrect`) — so every column powdr receives is either original or comes with a recipe. Input `derived_columns` are still carried through verbatim. The gap (emit + prove the `Reencode` hint, relax the spec to entailment-preservation) is documented in `docs/design/architecture.md`.

### Correctness
- `lake build` fully green, no `sorry`/`admit`/new axioms.
- `#print axioms leanrOptimizeJson` = `[propext, Classical.choice, Quot.sound]` (unchanged).
- No audited-surface change: `Spec.lean`, `OpenVmSemantics.lean`, `MemoryBus.lean`, `Leanr/Optimizer.lean`, `Basic.lean` untouched. The new code lives in the implementation layer (`JsonSerializer.lean`, `Implementation/Optimizer.lean`, `Ffi.lean`) and docs.
- The serializer carries a `#guard` derived-column serialize→parse→serialize round-trip.

### Benchmarks (see the powdr PR for full commands/tables)
- **Round-trip** (`autoprecompiles/tests/lean_ffi_roundtrip.rs`): `keccak_roundtrip`, `single_div_nondet_roundtrip`, and a new `derived_columns_roundtrip` (a synthetic `QuotientOrZero` hint survives the FFI byte-for-byte) all pass.
- **Columns** (release, `POWDR_USE_LEAN_OPTIMIZER=1`; counts columns/bus/constraints): keccak Rust `2021/1734/186` vs Lean `3646/7796/512`; ecrecover Rust `3730/2314/3114` vs Lean `4966/9058/3583`; single_div Rust `47/24/44` vs Lean `47/24/52`. The verified Lean optimizer currently trails the native Rust solver on large circuits — it is capped at 32 cleanup cycles (each substitutes few vars), whereas the native optimizer runs to fixpoint. On the small circuit it matches.
- **Keccak wall time** (release): Rust `17s` vs Lean FFI `560s`. The Lean work runs inside the profile-independent Lean static lib, so release speeds up only the native path (Task 1's debug 128s→17s) and the gap widens.
- **`reencodePass` cost of disabling** (Lean CLI, small fixtures): identical variable counts on most; on `apc_030` reencode-on saved 8 vars (69→61). So the witgen-safe path costs little on average.

### DIFF
--- a/Leanr.lean
+++ b/Leanr.lean
@@ -1,5 +1,6 @@
 -- Root of the `Leanr` library.
 import Leanr.Optimizer
 import Leanr.Implementation.JsonParser
+import Leanr.Implementation.JsonSerializer
 import Leanr.Utils.Dsl
 import Leanr.Utils.Size
--- a/Leanr/Ffi.lean
+++ b/Leanr/Ffi.lean
@@ -13,22 +13,39 @@ result back into a powdr `SymbolicMachine` JSON string.
 On any parse error the function returns a `{"error": "..."}` object rather than a machine, so the
 Rust side can surface a clear message instead of an opaque deserialization failure. The function
 never throws across the FFI boundary.
+
+## Witgen safety
+
+The end-to-end path (powdr witgen consumes the optimized circuit) requires that every witness
+column can actually be computed. The audited `openVmOptimizer` runs `reencodePass`, which invents
+fresh boolean columns whose values powdr's witgen can only fill from a `DerivedVariable` hint —
+and the audited spec (`optimizerPreservesDerivedColumns`) forbids the optimizer from emitting such
+hints (it must carry derived columns verbatim; see `docs/design/architecture.md`). So this entry
+point uses `optimizerWithBusFactsNoReencode` (the same pipeline without `reencodePass`): still a
+verified refinement, and it never produces a hint-less witness column. Input `derived_columns` are
+carried through verbatim, so any hints powdr already attached survive the round-trip.
 -/
 
 set_option autoImplicit false
 
-open Leanr.OpenVM (openVmOptimizer babyBear)
+open Leanr.OpenVM (babyBear openVmFacts BusMap)
+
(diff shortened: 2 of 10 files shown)


leonardoalt:
Superseded: the FFI + derived-columns path is now PR #53 (merged), #60 (merged), and #63. Closing this combine PR.

PR #60, opened 2026-07-08
Derived columns: emit ComputationMethods for optimizer-introduced columns

georgwiese:
Adds **derived columns** to the optimizer: a way to say *how witness generation computes the
columns the optimizer introduces*, so a verified optimization that adds columns (like the
re-encoder) can be used end-to-end. Mirrors powdr's `ComputationMethod<T, E>`.

## What to review — the audited spec (`Leanr/Spec.lean`)

- **`ComputationMethod`** — `const` / `quotientOrZero` / `ifEqZero`, mirroring powdr's
  `Constant` / `QuotientOrZero` / `IfEqZero`. `eval` matches powdr's `evaluate_computation_method`
  (`quotientOrZero num den` is `num / den`, or `0` when `den = 0`; needs no primality since `ZMod`
  has `Inv` for all `p`).
- **`Derivations := List (Variable × ComputationMethod)`** — the optimizer's extra output. The
  optimizer type is now `ConstraintSystem → ConstraintSystem × Derivations`.
- **`reconstructs` / `derivesWitness` / `impliesAdmissibleD` / `refinesD`** strengthen the
  completeness direction. The reproduced witness `env'` must `derivesWitness`:
  - every output variable **with** a powdr ID is an input column, present in the input with an
    **unchanged value**;
  - every output variable **without** a powdr ID is computed by a `ComputationMethod` in
    `Derivations` that reads **only input columns**.
- **`optimizerMaintainsCorrectness`** now bundles `refinesD` (+ invariant preservation + degree
  bound), threading the returned `Derivations`.

The audited theorem statements in `Leanr/Optimizer.lean` are unchanged in shape
(`optimizerWithBusFacts_maintainsCorrectness` and its `simpleOptimizer` / `openVmOptimizer`
instances) — they just instantiate the strengthened `optimizerMaintainsCorrectness`.

## One substantive design decision — input-columns-only

`reconstructs` requires each method to read **only genuine input (powdr-ID) columns**, and the
obligation is demanded only when the input circuit's columns all carry powdr IDs (the intended
shape of an exported circuit — a column with no powdr ID cannot be read from the input trace).
This is simpler than a general "read previously-derived columns in a valid order" form and
composes cleanly (the threading base case `cs.reconstructs [] env` holds vacuously on an
all-input-column input). Flagging it since it is the one place the spec makes a real choice.

**`ComputationMethod` was sufficient** — powdr's enum did not need extending.

## Implementation (no audit surface)

- Every existing pass threads `Derivations`; passes that add no columns ship `[]`. The framework's
  `PassCorrect` now carries the reconstruction obligation, so a pass **cannot** introduce a column
  without also saying how to compute it.
- The re-encoder emits, for each fresh boolean bit, a decision-tree `ComputationMethod` (`ifEqZero`
  over the group's joint values) that inverts the interpolation, and proves it matches the forward
  witness by construction.
- `reencodeOut_vars_subset` proves the output-variable frame by construction, replacing two
  `cs.vars` scans that had regressed runtime on the large (~5k-variable) benchmark cases (one
  cleanup cycle on `apc_005`: 120s+ → 63s).

## Proof integrity

- `lake build` green; no `sorry` / `admit` / `native_decide` / new axioms.
- `#print axioms` on both master theorems shows only `[propext, Classical.choice, Quot.sound]`.
- The optimizer's circuit transformation is behavior-preserving vs. the pre-derived-columns
  baseline (verified per case on the small cases, e.g. `apc_001` 134→42, `apc_003` 468→133); the
  feature only adds the `Derivations` output. Full-benchmark aggregate is being confirmed.

https://claude.ai/code/session_01DaV2Btmhk3brMJkusD2g2J

---
_Generated by [Claude Code](https://claude.ai/code/session_01DaV2Btmhk3brMJkusD2g2J)_

### DIFF
--- a/Leanr/Ffi.lean
+++ b/Leanr/Ffi.lean
@@ -30,5 +30,7 @@ def leanrOptimizeJson (input : String) : String :=
   match parseJsonSystem (p := babyBear) input with
   | .error err => "{\"error\":" ++ escapeJson err ++ "}"
   | .ok (cs, busMap) =>
-    let optimized := openVmOptimizer busMap.toBusMap cs
+    -- `.1` is the optimized circuit; `.2` are the `Derivations` for witness generation, not yet
+    -- threaded through the JSON serialization (would need a matching powdr-side format).
+    let optimized := (openVmOptimizer busMap.toBusMap cs).1
     Leanr.Serialize.serializeSystem optimized
--- a/Leanr/Spec.lean
+++ b/Leanr/Spec.lean
@@ -46,7 +46,9 @@ def Expression.vars : Expression p → List Variable
 --------- Computation Methods ---------
 
 /-- A method for computing a *derived* variable's value from other variables, mirroring powdr's
-    `ComputationMethod`. `quotientOrZero num den` is `num / den` in the field, or `0` when
+    `ComputationMethod`. For newly introduced variables, this is interpreted by powdr's witness
+    generator.
+    `quotientOrZero num den` is `num / den` in the field, or `0` when
     `den = 0`; `ifEqZero cond thenM elseM` picks `thenM` when `cond` evaluates to `0`, else `elseM`. -/
 inductive ComputationMethod (p : ℕ) where
   | const (c : ZMod p)
@@ -154,6 +156,34 @@ def ConstraintSystem.sideEffects (cs : ConstraintSystem p)
       let m := bi.eval env
       ((m.busId, m.payload), m.multiplicity))
 
+--------- Derived variables ---------
+
+/-- The `ComputationMethod` witness generation uses for `v`: the **last** one `ds` lists for it
+    (later derivations override earlier ones), or `none` if `v` is not derived. -/
+def Derivations.methodFor : Derivations p → Variable → Option (ComputationMethod p)
+  | [], _ => none
+  | (u, cm) :: rest, v =>
+      (Derivations.methodFor rest v).orElse (fun _ => if u = v then some cm else none)
+
+/-- Derived-variable reconstruction under `e`: every no-powdr-ID variable of `cs` is computed by
+    the method `ds` uses for it, reading only input (powdr-ID) columns. -/
+def ConstraintSystem.reconstructs (cs : ConstraintSystem p) (ds : Derivations p)
+    (e : Variable → ZMod p) : Prop :=
+  ∀ v ∈ cs.vars, v.powdrId? = none →
+    ∃ cm, Derivations.methodFor ds v = some cm ∧ (∀ x ∈ cm.vars, x.powdrId?.isSome) ∧ cm.eval e = e v
+
+/-- How the completeness witness `env'` of an output `out` is obtained from an assignment `env` of
+    the input `inp`, via derivations `ds`: every variable of `out` that carries a powdr ID is an
+    input column, present in `inp` with an unchanged value; every derived variable (no powdr ID) is
+    computed by a method in `ds` reading only input columns. This is exactly the data witness
+    generation needs to extend an input trace to an output trace. -/
+def ConstraintSystem.derivesWitness (out inp : ConstraintSystem p) (ds : Derivations p)
+    (env env' : Variable → ZMod p) : Prop :=
+  (∀ v ∈ out.vars, v.powdrId?.isSome → v ∈ inp.vars ∧ env' v = env v) ∧
+  out.reconstructs ds env'
+
+--------- Constraint system implications ---------
+
 /-- Whether a given assignment is admissible under the bus semantics. -/
 def ConstraintSystem.admissible (s : ConstraintSystem p) (busSemantics : BusSemantics p)
     (env : Variable → ZMod p) : Prop :=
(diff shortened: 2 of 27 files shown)


Leanr/Spec.lean
```diff
@@ -36,6 +36,41 @@ def Expression.degree : Expression p → Nat
   | .add e1 e2 => max e1.degree e2.degree
   | .mul e1 e2 => e1.degree + e2.degree
 
+/-- The variables occurring in an expression. -/
+def Expression.vars : Expression p → List Variable
+  | .const _ => []
+  | .var x => [x]
+  | .add e1 e2 => e1.vars ++ e2.vars
+  | .mul e1 e2 => e1.vars ++ e2.vars
```
leonardoalt, line 44: Why was this moved from `Size.lean`?

Leanr/Spec.lean
```diff
@@ -146,25 +187,60 @@ def ConstraintSystem.implies (self other : ConstraintSystem p) (busSemantics : B
     ∃ env', other.satisfies busSemantics env' ∧
       self.sideEffects busSemantics env ≈ other.sideEffects busSemantics env'
 
+--------- Derived variables ---------
+
+/-- The `ComputationMethod` witness generation uses for `v`: the **last** one `ds` lists for it
+    (later derivations override earlier ones), or `none` if `v` is not derived. The tail (later
```
leonardoalt: Does this ever happen, that a var has multiple derivations?

Leanr/Spec.lean
```diff
@@ -146,25 +187,60 @@ def ConstraintSystem.implies (self other : ConstraintSystem p) (busSemantics : B
...
+    ∃ cm, Derivations.methodFor ds v = some cm ∧ (∀ x ∈ cm.vars, x.powdrId?.isSome) ∧ cm.eval e = e v
+
+/-- How the completeness witness `env'` of an output `out` is obtained from an assignment `env` of
+    the input `inp`, via derivations `ds`: every variable of `out` that carries a powdr ID is an
+    input column, present in `inp` with an unchanged value; every derived variable (no powdr ID) is
+    computed by a method in `ds` reading only input columns. This is exactly the data witness
+    generation needs to extend an input trace to an output trace. -/
+def ConstraintSystem.derivesWitness (out inp : ConstraintSystem p) (ds : Derivations p)
+    (env env' : Variable → ZMod p) : Prop :=
+  (∀ v ∈ out.vars, v.powdrId?.isSome → v ∈ inp.vars ∧ env' v = env v) ∧
```
leonardoalt: Can it happen that a derived var depends on another derived var? I guess nothing really prevents that? Looks like this spec assumes that doesn't happen.

Leanr/Spec.lean
georgwiese: I think to keep `Spec.lean` self-contained. makes sense to me. `Size.lean` is not part of the *correctness* auditing surface, because it doesn't influence correctness (only how success is measured).

Leanr/Spec.lean
georgwiese: I don't think it should happen in practice, but the representation allows for it.

Leanr/Spec.lean
georgwiese: I think it *enforces* it not to happen. So it's stricter than it needs to be, but seems like it's fine for the current optimization passes, and I'm not aware of one that would need it to be less strict.

Leanr/Ffi.lean
```diff
@@ -30,5 +30,7 @@ def leanrOptimizeJson (input : String) : String :=
   match parseJsonSystem (p := babyBear) input with
   | .error err => "{\"error\":" ++ escapeJson err ++ "}"
   | .ok (cs, busMap) =>
-    let optimized := openVmOptimizer busMap.toBusMap cs
+    -- `.1` is the optimized circuit; `.2` are the `Derivations` for witness generation, not yet
+    -- threaded through the JSON serialization (would need a matching powdr-side format).
+    let optimized := (openVmOptimizer busMap.toBusMap cs).1
```
georgwiese, line 35: @leonardoalt Seems like we need a follow-up PR.

leonardoalt: I can do that together with the powdr side

Leanr/Spec.lean
leonardoalt: Where does it enforce it?

Leanr/Spec.lean
georgwiese: Note that this last-write wins is what would matter in this case, because [that's what the trace generator does](https://github.com/powdr-labs/powdr/blob/14b3e3053062c8bb510499cbe07763d5065e1f01/openvm/src/powdr_extension/trace_generator/cpu/mod.rs#L186-L195).

Leanr/Spec.lean
georgwiese: https://github.com/powdr-labs/leanr/blob/3186bdd169eb06cc7264ce7e3563fd02b7201786/Leanr/Spec.lean#L173

leonardoalt: That doesn't enforce the non-existence of a derived that depends on derived

PR #68, opened 2026-07-08
Thread derivations through passes to keep methods valid

georgwiese:
## Summary

Fixes leanr #64: derived-variable computation methods now reference only columns present in the optimized circuit, making them realizable by powdr's witness generator.

Previously, the re-encoding pass emitted `ComputationMethod`s as decision trees over the original group columns `xs` — the very columns it substituted out of the circuit. Later passes that eliminated variables did not update the accumulated derivations, so exported methods referenced removed columns, causing powdr's witness generator to panic.

The fix implements Georg's guidance: "whenever a variable gets removed, scan the derivations and replace it there as well." The pass framework now **threads** derivations (`dsIn ↦ dsOut`) rather than concatenating them, and variable-eliminating passes substitute the removed variable inside every accumulated method.

## Key Changes

**Spec (`Leanr/Spec.lean`, audited):**
- `ConstraintSystem.reconstructs` now requires each derived variable's method to read only variables **present in the output system** (`∀ x ∈ cm.vars, x ∈ cs.vars`), not just powdr-ID columns. This ensures methods are realizable on the optimized circuit alone.

**Framework (`Leanr/Implementation/OptimizerPasses/Basic.lean`, `FactPass.lean`):**
- `PassCorrect` now threads concrete derivations: `PassCorrect cs out dsIn dsOut bs` instead of quantifying over arbitrary `dsIn` with `dsLocal` appended.
- `PassCorrect.refl` and `PassCorrect.andThen` updated to thread derivations through composition.
- `PassCorrect.ofEnvEq` (for non-variable-introducing passes) requires empty incoming derivations and produces empty outgoing ones; variable-removing passes use `guardEmpty` to enforce this.

**Substitution (`Leanr/Implementation/OptimizerPasses/SubstMap.lean`):**
- New `ComputationMethod.substF` substitutes a variable-to-expression map inside every method.
- `ConstraintSystem.substF_correct_D`: the derivations-carrying version of `substF_correct`, threading incoming derivations through with substitution applied.
- Helper theorems (`Expression.substF_vars'`, `ComputationMethod.substF_vars'`) characterize which variables survive substitution.

**Re-encoding pass (`Leanr/Implementation/OptimizerPasses/Reencode.lean`):**
- `ConstraintSystem.reencode_correct_D`: updated to thread `dIn ↦ dOut` instead of quantifying over `dsIn`.
- `checkReencode_sound_D`: emits derivations as `(dIn ++ fresh_bits).map (fun x => (x.1, x.2.substF (groupSubst xs hm)))`, substituting the removed group columns into every method.
- New theorems (`Expression.substF_vars_out`, `ComputationMethod.substF_vars_out`, `groupRewrite_preserves_notmem`, `reencodeOut_vars_rev`, `reencodeOut_bit_mem`) prove that substituted methods reference only surviving columns and fresh bits.

**Other passes:**
- `MemoryUnify.addConstraints_correct_D`: threading version (no variables removed, so `dIn ↦ dIn`).
- All other passes updated to accept `dsIn` parameter and thread it through (most use `guardEmpty` since they don't eliminate variables).

**Diagnostics (`Main.lean`):**
- New `cmdCheckDerivs` command confirms every emitted derivation's method references only columns in the optimized circuit (should report 0 bad references).

**Documentation (`docs/log.md`, `docs/design/architecture.md`):**
- Explains the spec fix, the threading mechanism, and how variable-eliminating passes keep derivations valid by substituting removed variables.

## Implementation Details

- Substitution is applied lazily: methods are substituted only when emitted, not during intermediate passes.
- The threading is transparent to passes that don't eliminate variables; they simply pass derivations through unchanged.
- The re-encoder's `bitCM` methods (decision trees over `xs`) are substit

https://claude.ai/code/session_0158sKiZC3yGSMVxknwNkLBf

### DIFF
--- a/Leanr/Implementation/Optimizer.lean
+++ b/Leanr/Implementation/Optimizer.lean
@@ -67,6 +67,13 @@ theorem cleanupCycle_respectsDeg : RespectsDeg (cleanupCycle (p := p)) := by
   repeat' apply VerifiedPassW.andThen_respectsDeg
   all_goals exact VerifiedPassW.guardDegree_respectsDeg _
 
+/-- The cleanup cycle runs to a fixpoint. Derivations are **threaded**: the derivation-producing
+    pass (`reencodePass`) runs inside the cycle, and the variable-eliminating passes that follow it
+    in later cycles (`gaussElimPass`, `domainBatchPass`, `busUnifyPass`) substitute the removed
+    variables inside the accumulated derivations, so every emitted method keeps referencing only
+    columns present in the optimized circuit (leanr #64). The remaining, non-eliminating passes see
+    only empty derivations (they run before the first re-encode) and so apply unconditionally
+    (`guardEmpty`). -/
 def pipeline : VerifiedPassW p :=
   constantFoldPass.withFacts.guardDegree
     |>.andThen (iterateToFixpoint cleanupCycle)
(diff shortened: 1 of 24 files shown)


georgwiese:
not needed

PR #70, opened 2026-07-08
Improve spec readability

georgwiese:


Leanr/Spec.lean
```diff
@@ -165,20 +165,18 @@ def Derivations.methodFor : Derivations p → Variable → Option (ComputationMe
...
+    records for it, read from the input columns. This is what powdr runs to fill the optimized
+    circuit's columns from an input trace. -/
+def Derivations.witgen (ds : Derivations p) (inputEnv : Variable → ZMod p) : Variable → ZMod p :=
+  fun v =>
+    match v.powdrId? with
+    | some _ => inputEnv v
+    | none =>
+      match Derivations.methodFor ds v with
+      | some cm => cm.eval inputEnv
+      | none => inputEnv v
```
leonardoalt, line 192: this looks odd?

Leanr/Spec.lean
```diff
@@ -203,39 +201,28 @@ def ConstraintSystem.guaranteesInvariants (s : ConstraintSystem p) (busSemantics
...
-
-/-- Whether `self` is a valid **optimization** of `other` under a given bus semantics:
-    * **sound** — `self.implies other`: A satisfying assignment of `self` implies that there exists
-      a satisfying assignment of `other` with the same side effects.;
-    * **complete for admissible executions** — `other.impliesAdmissible self`: every *admissible*
-      (real-trace) satisfying assignment of `other` is reproduced by `self`. -/
-def ConstraintSystem.refines (self other : ConstraintSystem p) (busSemantics : BusSemantics p)
-    (ds : Derivations p) : Prop :=
-  self.implies other busSemantics ∧ other.impliesAdmissible self busSemantics ds
+  (∀ v ∈ originalCS.vars, v.powdrId?.isSome) →
```
leonardoalt, line 234: Shouldn't all original vars have an id?

Leanr/Spec.lean
```diff
@@ -203,39 +201,28 @@ def ConstraintSystem.guaranteesInvariants (s : ConstraintSystem p) (busSemantics
...
+  ∀ env, originalCS.admissible busSemantics env → originalCS.satisfies busSemantics env →
+    let env' := Derivations.witgen ds env
+    optimizedCS.satisfies busSemantics env' ∧ optimizedCS.admissible busSemantics env' ∧
```
leonardoalt, line 238: I actually find this block a bit harder to read than the old one

Leanr/Implementation/OptimizerPasses/Basic.lean
```diff
@@ -40,6 +41,23 @@ theorem BusState.equiv_symm {s t : BusState p} (h : s ≈ t) : t ≈ s :=
 theorem BusState.equiv_trans {s t u : BusState p} (h1 : s ≈ t) (h2 : t ≈ u) : s ≈ u :=
   fun message => (h1 message).trans (h2 message)
 
+/-- Soundness half of a replacement (glue, formerly in `Spec.lean`): every satisfying assignment of
```
leonardoalt, line 44: "soundness half of a replacement" sounds weird, what does it mean?

Leanr/Implementation/Optimizer.lean
```diff
@@ -94,15 +94,90 @@ def optimizerWithBusFacts {bs : BusSemantics p} (facts : BusFacts p bs)
...
-    because nontrivial facts are tied to one semantics. -/
+/-! ## Evaluation depends only on a system's variables
+
+Two assignments that agree on `cs.vars` are interchangeable for `satisfies`, `admissible`, and
+`sideEffects`. These lift the per-expression `Expression.eval_congr` / `BusInteraction.eval_congr`
+(proved among the passes) to the system level; the completeness bridge below uses them to swap the
+abstract per-pass witness for the concrete `witgen` output the spec now names. -/
+
+/-- A bus interaction of `cs` evaluates the same under assignments agreeing on `cs.vars`. -/
+theorem ConstraintSystem.busEval_congr {cs : ConstraintSystem p} {f g : Variable → ZMod p}
```
leonardoalt, line 105: Why do we need all these new theorems and why are the proofs so big?

Leanr/Spec.lean
georgwiese: I agree. It is not clear to me what happens when `inputEnv` is queried for a variable that's not in the input constraint system. Will have a look tomorrow. But I like the idea of just implementing the witgen algorithm here.

Leanr/Spec.lean
georgwiese: Yes, this should always be true in practice. But I don't think the statement below is provable without it.

Leanr/Spec.lean
georgwiese: Really, why? The difference is that previously there is a `∃ env', ... ∧ ((∀ v ∈ self.vars, v.powdrId?.isSome) → other.derivesWitnessFrom self ds env env')`. I find it a lot more intuitive to say that `env'` is deterministically computed from the input env and derivations.

Leanr/Implementation/Optimizer.lean
georgwiese: I have no idea honestly

Leanr/Implementation/OptimizerPasses/Basic.lean
georgwiese: This is `Implementation`, I don't look at it. I think what's happening is that large part of the codebase still has the old, slightly different concepts and it wrote some adapter theorems.

leonardoalt: I mean the grammar sounds weird

Leanr/Spec.lean
georgwiese: I added `Derivations.cover` to make sure that it can only access variables from `inputEnv` that actually appear in the constraint system.

Leanr/Spec.lean
leonardoalt: yea it's fine, it somehow just took me longer to understand this complete def than the previous

Leanr/Implementation/Optimizer.lean
leonardoalt: fair enough

PR #71, opened 2026-07-08
Add domain-constant subexpression folding pass

georgwiese:
## Summary

Implements `domainFoldPass`, a new optimization pass that folds subexpressions constant on a variable group's constrained domain to their constant values, without introducing bits or modifying the group's variables. This recovers the ~87% indirect effectiveness (register/memory chaining collapse) that `reencode` achieves as a side effect, making the mechanism explicit and powdr-aligned.

## Changes

- **New file: `Leanr/Implementation/OptimizerPasses/DomainFold.lean`** (427 lines)
  - `constOnSurvs`: checks if an expression evaluates to the same constant on all domain survivors
  - `foldRewrite`: recursively replaces maximal wholly-in-group compound subexpressions that are constant on survivors
  - `foldOut`: applies the rewrite to non-covered constraints and bus interactions, keeping covered (domain-pinning) constraints verbatim
  - `foldStep` / `foldLoop`: processes candidate variable groups (2–8 vars, same targets as `reencode`)
  - `domainFoldPass`: the top-level verified pass
  - Correctness via `PassCorrect.ofEnvEq`: the folded equation `e − c = 0` is entailed by kept covered constraints, so any satisfying assignment pins the group to a survivor where the rewrite is the identity

- **Updated: `Leanr/Implementation/Optimizer.lean`**
  - Import `DomainFold`
  - Insert `domainFoldPass` **before** `busUnifyPass` in `cleanupCycle` (so folded constants unblock `addrConstsNeq` chaining in the same cycle)

- **Updated: `docs/log.md`**
  - Entry 47: investigation findings (87% indirect / 13% direct split of `reencode`'s effectiveness; powdr's 1808 on apc_005 via range/domain simplification)
  - Entry 48: this pass's design, measurement (fold-only: 3619 → 1939 vars; fold+reencode: 1683, byte-identical to reencode-only)

- **Updated: `docs/ideas.md`**
  - Added decision point: option B (drop `reencode`, keep fold-only at 1939 vars) vs option A (keep both, current config)

## Implementation Details

- **Soundness**: Folding is a pure rewrite (`env' = env`, no new variables, no derivations). The covered constraints (group-local domain pinning) are kept verbatim, so any assignment satisfying the output must satisfy them and thus pin the group to a survivor, under which `foldRewrite_agree_covered` proves the rewrite is evaluation-preserving.
- **Efficiency gate**: `systemHasFoldable` Bool check prevents no-op passes; folding only lowers degree so the guard never blocks valid folds.
- **Shared infrastructure**: Reuses `groupDoms` / `groupSurvivors` from `reencode` (requires prime `p`, gated at runtime).
- **Proof structure**: Agreement theorems (`foldRewrite_agree`, `foldRewrite_agree_covered`) establish the identity on survivor-matching environments; `foldOut_correct` composes them into `PassCorrect.ofEnvEq`.

All proofs are complete; `lake build` green and `check-proof-integrity.sh` passes (no `sorry`, `admit`, `axiom`, or `native_decide`).

https://claude.ai/code/session_012wUkMJzbfTehax6XVKWDjb

### DIFF
--- a/Leanr/Implementation/Optimizer.lean
+++ b/Leanr/Implementation/Optimizer.lean
@@ -16,6 +16,7 @@ import Leanr.Implementation.OptimizerPasses.BusUnify
 import Leanr.Implementation.OptimizerPasses.BusPairCancel
 import Leanr.Implementation.OptimizerPasses.DisconnectedComponent
 import Leanr.Implementation.OptimizerPasses.Reencode
+import Leanr.Implementation.OptimizerPasses.DomainFold
 
 set_option autoImplicit false
 
@@ -60,6 +61,7 @@ def cleanupCycle : VerifiedPassW p :=
     |>.andThen trivialConstraintDropPass.withFacts.guardDegree
     |>.andThen zeroMultBusDropPass.withFacts.guardDegree
     |>.andThen tautoBusDropPass.withFacts.guardDegree
+    |>.andThen domainFoldPass.withFacts.guardDegree
     |>.andThen busUnifyPass.guardDegree
     |>.andThen (VerifiedPassW.guardDegree (iterateToFixpoint busPairCancelPass))
     |>.andThen disconnectedComponentPass.withFacts.guardDegree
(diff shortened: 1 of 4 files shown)


georgwiese:
No improvements, but also not slower. Seems to improve robustness, merging

PR #77, opened 2026-07-09
Optimizer runtime: eliminate per-constraint rescans in hintCollapse and reencode

georgwiese:
## Summary

Pure performance optimization (entry 53) that eliminates redundant full-system rescans in `hintCollapse` and `reencode` passes. All changes are output-preserving — effectiveness is unchanged, only wall-clock optimizer time drops. Verified on 13 benchmark cases (apc_001/003/005/006/008/010/014/028/047/056/069/092/100): variable/constraint/bus counts are identical on every case.

## Key Changes

**`HintCollapse.lean`:**
- Hoist `BoundsMap.build` and a new `busVars` hash set (all variables in any bus interaction) to pass level, built once per invocation instead of once per constraint
- Gate `occursOnlyInTarget` behind `!busVars.contains v` — variables in bus interactions can never pass the check, so this short-circuits the full-system scan without changing the filter's value
- Replace `d ∉ ·.vars` membership checks with allocation-free `Expression.mentions` (early-exit tree walk) in `occursOnlyInTarget`
- Convert the certificate conjunction into a short-circuiting `&&` chain so conjuncts after `2 ≤ D.length` (which fails on ~99% of constraints) are never evaluated
- Update proofs to consume the same certificates via `Bool.and_eq_true`/`decide_eq_true_eq` decomposition and new `mentions_false_not_mem_vars` lemma

**`Reencode.lean`:**
- Move the `xs.all (· ∈ cs.vars)` membership check after `buildReencode` succeeds (pure branch reorder — all failure paths return the same identity result)
- Bind `cs.vars` once with a `let` to avoid materializing the ~10⁴-entry occurrence list for every candidate group

**`Main.lean`:**
- Sync the stale `leanr profile` pass list to current `cleanupCycle` (add `zeroRegister`, `hintCollapse`, `domainFold`, `busPairCancel`, `bytePack`)

**`docs/log.md` and `docs/ideas.md`:**
- Document the performance analysis, fixes, and measured impact
- Update bottleneck notes: `busPairCancel` is now a measured bottleneck (24 s on apc_005)

## Impact

**Solo runs (same machine, output identical):**
- apc_005: **280.5 s → 75.3 s (3.7×)**
- apc_100: **320.0 s → 155.7 s (2.1×)**
- apc_010: 14.6 s → 8.8 s (1.7×)
- Per-pass on apc_005: `hintCollapse` **182.9 s → 0.2 s (~900×)**, `reencode` **37.7 s → 5.2 s (7.3×)**

Small register-only cases unchanged (they never hit the hot paths).

## Implementation Details

The fixes are output-preserving because:
1. `busVars` short-circuit only skips the scan when the result would be `false` anyway
2. `mentions` is equivalent to `∉ ·.vars` (proven via `mentions_false_not_mem_vars`)
3. Moving the `reencode` membership check after `buildReencode` is a pure branch reorder
4. The certificate `&&` chain is logically equivalent to the original conjunction — it just short-circuits earlier

Correctness axioms remain `{propext, Classical.choice, Quot.sound}`; `lake build` passes and `check-proof-integrity.sh` passes.

https://claude.ai/code/session_01Wm6YUkLPPZJj8uALRyPZCq

### DIFF
--- a/Leanr/Implementation/OptimizerPasses/HintCollapse.lean
+++ b/Leanr/Implementation/OptimizerPasses/HintCollapse.lean
@@ -470,28 +470,33 @@ theorem collapse_correct [Fact p.Prime] (cs : ConstraintSystem p) (bs : BusSeman
 
 /-! ## Detection: find a collapsible reciprocal-witness group and discharge the facts -/
 
-/-- A witness variable `d` occurs (in the whole system) only in the target constraint `E`. -/
+/-- A witness variable `d` occurs (in the whole system) only in the target constraint `E`.
+    Decided with the allocation-free `Expression.mentions` (early-exit tree walk) rather than
+    `d ∉ ·.vars`, which would materialize every expression's variable list once per `(E, d)`
+    pair — this check runs inside the per-constraint candidate scan. -/
 def occursOnlyInTarget (cs : ConstraintSystem p) (E : Expression p) (d : Variable) : Bool :=
-  (cs.algebraicConstraints.all (fun c => decide (c = E) || decide (d ∉ c.vars))) &&
+  (cs.algebraicConstraints.all (fun c => decide (c = E) || !(c.mentions d))) &&
   (cs.busInteractions.all (fun bi =>
-    decide (d ∉ bi.multiplicity.vars) && bi.payload.all (fun e => decide (d ∉ e.vars))))
+    !(bi.multiplicity.mentions d) && bi.payload.all (fun e => !(e.mentions d))))
 
 theorem occursOnlyInTarget_constr {cs : ConstraintSystem p} {E : Expression p} {d : Variable}
     (h : occursOnlyInTarget cs E d = true) : ∀ c ∈ cs.algebraicConstraints, d ∈ c.vars → c = E := by
   intro c hc hdc
   simp only [occursOnlyInTarget, Bool.and_eq_true, List.all_eq_true] at h
   have hc' := h.1 c hc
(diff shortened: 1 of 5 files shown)


georgwiese:
Nice! Faster and slightly better.

Main:
```
leanr compare: 100%|██████████████| 100/100 [04:20<00:00,  2.61s/case]

=== openvm-eth: leanr vs powdr over 100 cases ===
effectiveness = size before / size after (larger is better); priority: variables > bus interactions > constraints
  measure           leanr (agg / geo)         powdr (agg / geo)         diff (agg)
  variables         4.036x / 3.442x           4.092x / 3.787x           -0.055x
  bus interactions  2.916x / 2.426x           3.480x / 2.822x           -0.564x
  constraints       8.709x / 9.251x           5.853x / 10.311x          +2.856x
per-case (by variables): leanr wins 14, loses 84, ties 2
````

This branch
```
leanr compare: 100%|██████████████| 100/100 [03:44<00:00,  2.24s/case]

=== openvm-eth: leanr vs powdr over 100 cases ===
effectiveness = size before / size after (larger is better); priority: variables > bus interactions > constraints
  measure           leanr (agg / geo)         powdr (agg / geo)         diff (agg)
  variables         4.082x / 3.605x           4.092x / 3.787x           -0.010x
  bus interactions  2.922x / 2.440x           3.480x / 2.822x           -0.557x
  constraints       8.801x / 9.918x           5.853x / 10.311x          +2.948x
```

PR #80, opened 2026-07-10
Speed up the optimizer 3.7-4.2x on the profiled cases (output unchanged)

pacheco:
Pure performance work on the three dominant passes; the optimizer's output is unchanged (verified — see below). Measured `main` vs this PR (solo runs, same machine):

| case | total | domainBatch | busPairCancel | reencode |
|---|---|---|---|---|
| apc_006 | **101.8s → 24.2s (4.2×)** | 67.9s → 5.6s (12.1×) | 18.4s → 10.5s (1.8×) | 9.7s → 2.9s (3.3×) |
| apc_036 | **47.2s → 12.7s (3.7×)** | 18.6s → 1.8s (10.6×) | 17.5s → 2.8s (6.2×) | 3.8s → 1.7s (2.2×) |

## How the costs were found

Per-stage instrumentation of each pass, then `perf record` + reading the generated C. The decisive finding for domainBatch/reencode: **every `+`/`*`/`decide (· = 0)` on `ZMod p` with a runtime `p` re-derives the whole `CommRing (ZMod p)` instance chain per expression node** — `perf` showed most of domainBatch in the allocator building and freeing instance records, plus ~10% in `ZMod.commRing` and its projections directly. For busPairCancel, stage timing showed ~90% of the pass in the receive search (`findMatchRecv`): every invocation re-probes every send against the whole remaining list with structural payload comparisons (51k probes in one cleanup iteration of apc_036), repeated once per dropped pair by the fixpoint loop. Other structural costs: domainBatch's box enumeration paid the full `box × items` evaluation even for targets that force nothing, and reencode rebuilt/re-evaluated its interpolation candidates ~3× per bit pattern for candidate groups the degree guard then rejects — every cleanup iteration again.

## The commits (each builds green on its own)

**domainBatch**
- drop domain-redundant constraints from the per-target enumeration (`constraintRedundant`)
- `scanInit`/`scanWith`: single-pass box scan keeping the candidates every survivor agrees on, **aborting as soon as none remain** — claiming nothing needs no certificate; a completed scan *is* the certificate (`scanInit_some`)
- `RangeCache`: one materialization of `(List.range bound).map cast` per distinct bound (provably equal to uncached, `interactionDomainC_fst`)
- `varsInF`/`containsFast`: `powdrId?`-first comparison in the covered-item scans (proven extensionally equal)
- `IExpr`/`compiledSurv`: covered items compiled to positional leaves per target, with `add`/`mul`/`decide` extracted from the instances **once** (`survivesAllCW_eq` ∘ `survivesAllC_eq` is all the certificates consume; uncompiled fallback if compilation fails — it cannot, for covered items)

**reencode**
- bind the substituted expression, its per-pattern values, and the folded interpolation once (`interpOfV`/`candSelect`); evaluate through `Expression.evalFast` (field ops hoisted per call, `evalFast_eq`); reuse the covered set for the survivor filter (`groupSurvivorsE`)
- `powdrId?`-first comparisons in `coveredBy`/`groupSubst`/`groupRewrite` and the freshness sweep (`Expression.mentionsF`)

**busPairCancel**
- hash-index the candidate receives once per invocation (`recvIndex`/`payloadHash`) and scan sends over an `Array`, resolving each probe by hash lookup + exact payload verification on hits (`firstMatchAt`); hash inequality proves payload inequality, so exactly the same first match is found in the same order
- run the region tests over the array index ranges and materialize the A/B/C lists only for the candidates they accept (28k matched candidates on apc_006, 1.4k accepted)
- try `checkCancel` with `checks := []` before computing `unjustifiedSlots` (its non-justification conjuncts are guaranteed by the scan's gates, so it passes iff every byte slot is justified) — the common fully-justified drop pays the byte-justification scan once instead of twice

The audited surface (`Spec.lean`, `OpenVmSemantics.lean`, `MemoryBus.lean`, `ApcOptimizer/Optimizer.lean`, `Basic.lean`) is untouched; every fast path carries an extensional-equality lemma consumed by the existing proofs, or is re-verified by the pass's own checked certificate. `lake build` green; correctness axioms stay `{propext, Classical.choice, Quot.sound}`; `check-proof-integrity.sh` passes. Details in `docs/log.md` entries 54–55.

## Effectiveness is unchanged

Verified by A/B runs (`vars/constraints/bus` counts) between binaries differing only in the changed pass files:

- **`main` vs the branch through the domainBatch/reencode commits**: identical on the log-entry-53 13-case list (apc_001/003/005/006/008/010/014/028/047/056/069/092/100 — the same verification set the previous perf PR used, including the apc_005/apc_100 heavyweights) plus 10 random cases (seed 42) — 20 distinct cases.
- **That tip vs the final busPairCancel commits**: identical on the 13-case list plus apc_036 (14/14). End-to-end (`main` vs the PR head), the 13-case list is therefore verified directly; the remaining sampled cases are covered by chaining the two verified hops.
- Supporting evidence: before the repo-rename rebase, the domainBatch/reencode series was also checked against its own baseline — full `run` output byte-identical on apc_006, exact stats on apc_001–010 (matching the recorded log references, e.g. apc_001–008 bus total 1894) and 10 further random cases. On the instrumented cases (apc_006, apc_036), the passes' internal decision counters (forced values, candidate groups found/built/checked/accepted, send-probe/match/drop counts per iteration) are identical before/after.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/BusPairCancel.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/BusPairCancel.lean
@@ -976,11 +976,23 @@ def findCancelGoIdx (cs : ConstraintSystem p) (bs : BusSemantics p) (facts : Bus
           -- Cheap region tests first: only an otherwise-valid candidate pays the byte
           -- justification scan (`checkCancel` re-verifies everything).
           if B.all (midRefuted shape busId S) && A.all (preRefuted shape busId S) then
-          -- Byte slots the remaining system does not justify are materialized as a single
-          -- explicit self-check on the byte-check bus (more than one would not shrink the bus
-          -- count and would stall the cancellation loop); when the justification cannot pass
-          -- (≥ 2 unjustified slots, or no byte-check bus), skip the certificate re-check —
-          -- the justification scan is the expensive part.
+          -- Try the certificate with no emitted checks first: every non-justification conjunct
+          -- of `checkCancel` is guaranteed by the scan's own gates, so it passes iff every
+          -- declared byte slot is justified — the same predicate `unjustifiedSlots` decides —
+          -- and the common fully-justified drop pays the justification scan **once**.
+          if hchk0 : checkCancel deep cs.algebraicConstraints bs facts shape busId slots
+              A S B R C [] = true then
+            if hsplit : cs.busInteractions = A ++ S :: B ++ R :: C then
+              some ⟨{ cs with busInteractions := A ++ B ++ C ++ [] }, [],
+                    checkCancel_sound cs bs facts hp1 deep hdeep busId shape hshape slots hslots
+                      A S B R C [] hsplit hchk0⟩
+            else next ()
+          else
+          -- Some slot is unjustified. Such slots are materialized as a single explicit
+          -- self-check on the byte-check bus (more than one would not shrink the bus count and
+          -- would stall the cancellation loop); when the justification cannot pass (≥ 2
+          -- unjustified slots, or no byte-check bus), skip the certificate re-check — the
+          -- justification scan is the expensive part.
           let unjust := unjustifiedSlots deep cs.algebraicConstraints bs facts (A ++ B ++ C)
             slots R
           let checks : List (BusInteraction (Expression p)) :=
@@ -989,7 +1001,7 @@ def findCancelGoIdx (cs : ConstraintSystem p) (bs : BusSemantics p) (facts : Bus
                 [{ busId := bcBus, multiplicity := .const 1,
                    payload := [e, e, .const 0, .const 1] }])
             | _, _ => []
-          if unjust.isEmpty || !checks.isEmpty then
+          if !checks.isEmpty then
           if hchk : checkCancel deep cs.algebraicConstraints bs facts shape busId slots
               A S B R C checks = true then
             if hsplit : cs.busInteractions = A ++ S :: B ++ R :: C then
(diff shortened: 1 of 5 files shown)


leonardoalt:
Follow-up: runtime is now benched in CI. #82 adds a **Runtime Bench** job that times every optimizer call (`apc-optimizer run`) and attributes time per pass (`apc-optimizer profile`, cumulative over all fixpoint iterations) across the full 100-case openvm-eth set, on every PR and push to main — results land in the job summary and as a sticky PR comment. Locally: `OpenVmBenchmarks/runtime_bench.py`.

Baseline on post-#80 main (`3e28381`, local machine, serial, single run per case):

| total (100 cases) | mean | median | max |
|---|---|---|---|
| 412.5 s | 4.1 s | 591 ms | 92.7 s (apc_037) |

| pass | cumulative | share |
|---|---|---|
| busPairCancel | 222.0 s | 50.7% |
| domainBatch | 90.3 s | 20.6% |
| reencode | 40.4 s | 9.2% |
| domainFold | 33.4 s | 7.6% |
| gauss | 16.2 s | 3.7% |
| (14 others) | 35.6 s | 8.2% |

So this PR's three targets now cost 80.5% of pass time set-wide, with busPairCancel the clear remaining hotspot: it's concentrated in the two slowest cases, apc_037 (90.9 s of its 112.7 s profile) and apc_071 — a natural target for a next runtime pass, which the CI job will now measure automatically.

PR #90, opened 2026-07-10
Memory-pointer duplicate cleanup: two-root limb unification, syntactic dedup, flag unification

pacheco:
## What

Three verified passes (log entries 57–59) closing the duplicate memory-pointer-decomposition gap:

1. **`rootPairUnifyPass`** (entry 57): each memory access pins its pointer limbs with a two-root carry constraint `(A + k·x)(A + δ + k·x) = 0` plus a range check below the root gap. Same-address accesses produce twin constraints with distinct limb variables; the bounded-integer argument (`rootPair_eq`: two values below `min(g.val, p − g.val)` cannot differ by the root gap `±g` over the integers) proves the limbs equal. Bounds come from raw range checks (`findVarBound`) and scaled slots `4⁻¹·(x − F)` via a constant-coefficient decomposition `Expression.splitAt` + flag-domain enumeration + `ZMod` no-wrap val arithmetic.
2. **`dedupPass`** (entry 58): the unification turns the eliminated decomposition's carry constraints and raw-slot range check into literal copies — a keep-first dedup of syntactically identical constraints and *stateless* interactions collects them (stateful duplicates are observable in `sideEffects` and kept).
3. **`flagUnifyPass`** (entry 59): the surviving scaled check pins the eliminated access's flag variables; both checks decompose the *same* shared limb as `x = m·u + W`, so the flag-polynomial values are equal by integer residue uniqueness, and a ≤32-point joint flag-box enumeration certifies componentwise variable equality.

## Effectiveness (full 100-case openvm-eth sweep, tip vs merge base)

| axis | before | after | powdr |
|---|---|---|---|
| **variables** | 4.082× / 3.605× | **4.286× / 3.655×** | 4.092× / 3.787× |
| bus interactions | 2.922× / 2.440× | **3.006× / 2.466×** | 3.480× / 2.822× |
| constraints | 8.801× / 9.918× | **9.854× / 10.214×** | 5.853× / 10.311× |

- **First aggregate variable lead over powdr** (+0.194×); per-case wins 15 → 17, no per-case regressions (10-case sample across size classes byte-identical at each commit).
- apc_005 / apc_044 / apc_067 per block: **1683 → 1491 vars (−192), 841 → 649 constraints, 829 → 765 bus interactions**; powdr keeps 1808 vars on these.
- apc_005 wall-clock ~18 s (after the log-entry-60 runtime commit; was ~30 s).

## Correctness

- No changes to the audited surface; all three passes discharge `PassCorrect` (via `substF_correct` / `ofEnvEq`); prime `p` re-checked at runtime where needed, as in `busPairCancelPass`.
- `lake build` green at every commit; `Scripts/check-proof-integrity.sh` passes; all three `maintainsCorrectness` theorems remain `{propext, Classical.choice, Quot.sound}`-only.

Follow-ups recorded in `docs/ideas.md`: the last per-pair flag component relates non-componentwise (needs a derived-variable interpolation `b := f(a₀, a₁)`, worth ~−64 vars and −64 bus per apc_005-class block), and cross-offset chaining.

### DIFF
--- a/ApcOptimizer/Implementation/Optimizer.lean
+++ b/ApcOptimizer/Implementation/Optimizer.lean
@@ -23,6 +23,7 @@ import ApcOptimizer.Implementation.OptimizerPasses.HintCollapse
 import ApcOptimizer.Implementation.OptimizerPasses.CarryBranch
 import ApcOptimizer.Implementation.OptimizerPasses.RootPairUnify
 import ApcOptimizer.Implementation.OptimizerPasses.Dedup
+import ApcOptimizer.Implementation.OptimizerPasses.FlagUnify
 
 set_option autoImplicit false
 
@@ -68,6 +69,7 @@ def cleanupCycle : VerifiedPassW p :=
     |>.andThen zeroRegisterPass.guardDegree
     |>.andThen hintCollapsePass.guardDegree
     |>.andThen rootPairUnifyPass.guardDegree
+    |>.andThen flagUnifyPass.guardDegree
     |>.andThen dedupPass.withFacts.guardDegree
     |>.andThen trivialConstraintDropPass.withFacts.guardDegree
     |>.andThen zeroMultBusDropPass.withFacts.guardDegree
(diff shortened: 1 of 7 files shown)


leonardoalt:
The overall runtime gets considerably slower, do we wanna fix it here directly?

pacheco:
its already fixed, the bench didn't run again

leonardoalt:
Sigh that's annoying

PR #112, opened 2026-07-13
Sum-of-squares collapse for is-equal comparison gadgets (idea 4)

georgwiese:
## What

Implements **idea 4** from `docs/ideas.md` (and removes it): collapse the `beq`/`bne` **is-equal** comparison gadget's per-limb `diff_inv_marker` reciprocal witnesses down to a single derived hint, matching what powdr keeps.

After the guard-flag folds, the gadget leaves one bilinear constraint

```
Σᵢ (aᵢ − bᵢ)·mᵢ + (cmp − 1) = 0
```

where the `mᵢ` (`diff_inv_marker` limbs) occur **only** here (pure witnesses) and the coefficients are byte-bounded **differences**. `diff_inv_marker` is the single largest per-case variable family where apc trailed powdr. The existing `hintCollapse` could not fire: its coefficients must be a *single* byte-bounded variable, and its plain-sum denominator `Σ(aᵢ − bᵢ)` can vanish without every difference vanishing (they cancel).

The fix — mirroring powdr's is-equal witness — is the **sum of squares**:

```
(Σᵢ (aᵢ − bᵢ)²)·inv + (cmp − 1) = 0,   inv = QuotientOrZero(1 − cmp, Σ …²)
```

dropping all markers (**−3 variables per gadget**). A zero sum of squares forces every difference to `0`: each `(aᵢ − bᵢ)²` is a non-negative integer `< 256²`, so the field sum cannot wrap, and — `p` prime, an integral domain — each square, hence each difference, is `0`, pinning `cmp = 1`.

> Note: idea 4 proposed a *byte-weighted* difference `Σ 256ⁱ(aᵢ − bᵢ)`. That overflows the 31-bit BabyBear field for 4 limbs (4 byte-differences carry ~36 bits), so it is neither sound nor what powdr does. The sum-of-squares form is the correct, overflow-free construction.

## How

- `HintCollapse.lean`'s `collapse_correct` is generalized to take the soundness reassignment `reasg env w` as a parameter, so **one** proven theorem now backs both collapse shapes:
  - **plain sum** (`tryOne`, unchanged): single-variable byte-bounded coefficients, `is-zero`/`seqz`;
  - **sum of squares** (`tryOneSq`, new): byte-bounded difference coefficients, `is-equal`, with `mᵢ := (aᵢ − bᵢ)·w` so `Σ(aᵢ − bᵢ)·mᵢ = (Σ(aᵢ − bᵢ)²)·w`.
- New detection recognizes each peeled coefficient as a `fold`-normalized difference of two `≤ 256`-bounded variables and discharges completeness via a squared-byte-difference value bound + the existing no-wrap sum lemmas + `mul_self_eq_zero`.
- **No audited-surface change** (`Spec.lean`, `OpenVmSemantics.lean`, `MemoryBus.lean`, `ApcOptimizer/Optimizer.lean`, `Basic.lean` untouched); correctness follows from the pass's own `PassCorrect`. `lake build` green; `check-proof-integrity.sh` passes (theorems depend only on `propext`/`Classical.choice`/`Quot.sound`).

## Impact

openvm-eth, full 100-case `benchmark.py` sweep, same-session baseline vs. new build:

| measure | baseline (agg / geo) | new (agg / geo) | powdr (agg / geo) |
|---|---|---|---|
| **variables** | 4.509× / 3.820× | **4.517× / 3.835×** | 4.092× / 3.787× |
| bus interactions | 3.401× / 2.705× | 3.401× / 2.705× (unchanged) | 3.480× / 2.822× |
| constraints | 10.590× / 11.578× | 10.590× / 11.578× (unchanged) | 5.853× / 10.311× |

- **16 cases collapse, each −3 variables** (total apc eth vars 28026 → 27978); **0 bus regressions, 0 constraint regressions**.
- Per-case standings by variables: **25 W / 42 L / 33 T → 27 W / 29 L / 44 T** (13 losses eliminated — 11 flip loss→tie, 2 loss→win).
- The collapse produces powdr's exact `Σ(aᵢ − bᵢ)²·inv + cmp − 1` form.

## Not covered

The *signed*/`sltu` comparison markers (`diff_marker`, `c_msb_f`, `b_msb_f`) are a different mechanism — those markers appear in a bus **multiplicity**, so they are not pure reciprocal witnesses and this collapse does not apply. Recorded as a distinct follow-up in `docs/ideas.md`. Logged as entry 75 in `docs/log.md`.

https://claude.ai/code/session_01LoG4ZxvHFWkAHdJrBFGE8Y

---
_Generated by [Claude Code](https://claude.ai/code/session_01LoG4ZxvHFWkAHdJrBFGE8Y)_


pacheco:
i think my agent came up with a different solution to the same problem in #114, since there was low performance hit there, i merged it. But it might be that the solution here is more generic? or maybe it could be improved to have a low perf hit too?

georgwiese:
Good call — #114 and this PR are the same sum-of-squares collapse. The difference is structural, and it turns out that's exactly where the perf comes from.

I've rebased this branch onto `main` and reworked it into a **consolidation**: instead of adding a second pass, it generalizes `hintCollapse` — `collapse_correct` now takes the soundness reassignment as a parameter, so one proven theorem backs both witness shapes (the linear `is-zero`/`seqz` case and the quadratic sum-of-squares `is-equal` case), and both are tried in a **single** `hintCollapse` scan that computes the per-constraint witness set once and shares it. This branch therefore **removes the standalone `EqCollapse.lean`** (−640 lines; net −280 for the PR).

On both of your questions:

- **More generic** — yes: one collapse framework and one `collapse_correct` theorem for both shapes, rather than a separate 640-line pass.
- **Lower perf hit** — actually lower than `main`: on `main` `eqCollapse` runs as its own pass *in addition to* `hintCollapse` every cleanup iteration (a second whole-system witness scan). Folding it in roughly **halves** the combined collapse-stage time and removes the outliers noted in entry 76:

  | case | `main` (`hintCollapse` + `eqCollapse`) | this branch (`hintCollapse` only) |
  |---|---|---|
  | apc_044 | 203 + 198 = 401 ms | **187 ms** |
  | apc_037 | 621 + 640 = 1261 ms | **631 ms** |
  | apc_006 | 411 + 423 = 834 ms | **411 ms** |

**Effectiveness is byte-identical to `main`** (the −48 eth / −3 keccak vars already landed via #114; this reproduces them exactly): 100-case sweep 4.518×/3.837× vars, 3.479×/2.753× bus, 27 W / 29 L / 44 T; `apc_072` 29 = powdr parity. Build + `check-proof-integrity.sh` green.

Since #114 is already merged, this is effectively a proposal to swap the standalone pass for the generalized one — same result, fewer passes, ~half the collapse runtime. Entirely your call whether that consolidation is worth taking; happy to close if you'd rather keep #114 as-is.

---

georgwiese:
Cool!

PR #118, opened 2026-07-14
Width-1 range check -> booleanity: -68 keccak / -89 eth bus, 0 regressions — eth bus agg flips to a lead over powdr (+ two measured dead-ends)

pacheco:
## What

Extends `zeroWidthRangePass` with a **width-1 arm**: a range check `[e, 1]` on the variable range checker is accepted iff `e < 2` — booleanity — so it converts to the degree-2 algebraic constraint `e·(e−1) = 0` and the interaction is dropped. A strict lexicographic win (bus ≻ constraints, and constraint headroom vs powdr is large); the booleanity also feeds the finite-domain passes. No new pass, no new `BusFacts` — the width-1 iff comes from the existing mult-generic `varRangeBus` fact.

Two guards worth noting:
- the equivalence's backward direction (`x·(x−1) = 0 → x < 2`) needs an integral domain, so the arm gates on `decide (Nat.Prime p)` per invocation (the `deep`/`hdeep` pattern);
- a **per-candidate degree gate** (`e.degree ≤ 1`) keeps every emitted constraint at degree ≤ 2, so the arm can never trip the whole-pass `guardDegree` revert.

## Effectiveness (per-case JSON A/B vs main = #119 `9c92008`, re-measured after rebase)

Per-case deltas are identical to the pre-rebase measurement vs #117 — the win is fully independent of the entry-78 byte-pair splitting.

- **keccak: bus 2027 → 1959 (−68)** — bus effectiveness 6.543× → **6.770×**; gap to powdr now +225 (memory already at parity via #117; the residual is the genuine-XOR representation ≈212 + ~13 misc). Constraints 120 → 188; vars 2025 unchanged; runtime flat.
- **openvm-eth: bus agg 3.479× → 3.497×** (geo 2.753× → 2.759×) — the powdr bus gap goes **−0.001 → +0.018: the last trailing aggregate axis flips to a lead** (apc 16638 vs powdr 16722 total bus); 4 cases improved (apc_100 −40, apc_037 −25, apc_038/051 −12), −89 bus / +90 constraints / **0 regressions**; variables byte-neutral (4.518×, W/L/T 27/29/44 unchanged).

Build green; `check-proof-integrity.sh` passes ({propext, Classical.choice, Quot.sound}-only).

## Also included: two docs-only negative results (log entries 79–80)

- **Constant limb-decomposition folding (ideas.md #1, the ranked-top variable idea): mechanism refuted by a faithful what-if.** The clean affine `Σ cᵢ·xᵢ = K` equations exist only in the raw input and net −2/−3/0 vars on the flagship cases; the real +14/case gap vs powdr lives in a bounded-payload digit idiom (~3 vars/case) and guarded two-root carry chains needing joint multi-constraint interval reasoning (an optimistic what-if claiming −54/−141/−154 traced to unsoundly pinning one-hot selector sums). ideas.md #1 re-scoped accordingly.
- **Degree-bounded witness inlining + per-candidate degree planning: measured zero corpus-wide.** A cascade-capable alternation what-if (inline + certified `boxRewrite` legalization, alternated with the full optimizer to a joint fixpoint) accepted **0 inlinings on all 100 eth cases and keccak** — the few constant-coefficient nonlinear pivots that exist are blocked by the bus *payload* bound (the variable is the bus-facing value), beyond any booleanity legalization. Recorded as a dead-end.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/ZeroWidthRange.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/ZeroWidthRange.lean
@@ -3,97 +3,153 @@ import ApcOptimizer.Implementation.OptimizerPasses.FlagFold
 
 set_option autoImplicit false
 
-/-! # Width-0 range-check → equality (C3)
+/-! # Width-0 / width-1 range-check → equality / booleanity (C3 + entry 80)
 
-powdr-exported blocks that compute effective addresses keep, on the variable range checker, a family
-of **width-0 range checks** `[expr, 0]` (multiplicity `1`). A 0-bit range check asserts
-`expr < 2⁰ = 1`, i.e. `expr = 0`; it is powdr/OpenVM's encoding of "this linear form is exactly
-zero" — used to pin an intermediate address/data limb to a combination of others (e.g.
-`-a__2 + b__3 = 0`, `7864320·(a__3 − b__0) = 0`). Because the optimizer's Gaussian elimination only
-consumes *algebraic* constraints, these equalities — being carried on the **bus** — are never used
-to eliminate a variable, so the intermediate limbs survive. powdr substitutes them away.
+powdr-exported blocks keep, on the variable range checker, two families of degenerate range
+checks (multiplicity `1`):
 
-This pass converts each such interaction into the algebraic constraint `expr = 0` and drops the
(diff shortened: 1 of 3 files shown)


pacheco:
superseeded by #122

PR #139, opened 2026-07-16
Add SP1 bus semantics; make the bus facts VM-neutral

georgwiese:
## Summary

Adds an **SP1** bus semantics alongside the existing OpenVM one, and generalizes the shared
`BusFacts` so the byte-check / XOR / memory optimizations are **fully VM-neutral** — the same passes
now fire on both the OpenVM bitwise-lookup bus and the SP1 byte-lookup bus, with no OpenVM-shaped
payloads baked into the facts.

Audited-surface changes are small and localized (see below); the bulk is non-audited
`Implementation/` work whose correctness is discharged by each pass's own `PassCorrect`.

## Audited surface (what to review)

- **`Sp1Semantics.lean`** (new) — the KoalaBear-field SP1 `BusSemantics`: `isStateful`,
  `violates` (byte bus AND/OR/XOR/U8Range/…, plus the 16-bit memory discipline), `breaksInvariant`,
  `admissible`, degree bound. SP1 sends the previous record and receives the new one.
- **`MemoryBus.lean`** — the memory discipline is now parameterized by a `MemoryBusDirection` enum
  (`sendThenReceive` for OpenVM, `receiveThenSend` for SP1) via `setNewMult`; `getPrevious`/`setNew`
  terminology; theorems moved out to `Implementation/MemoryBusDrop.lean`. (All three spec-review
  comments addressed and resolved.)
- **`OpenVmSemantics.lean`** — declares `direction := .sendThenReceive` explicitly.
- **`Optimizer.lean`** — adds `sp1Optimizer` + `sp1Optimizer_maintainsCorrectness` next to the
  OpenVM instance, both instances of the existing master theorem.

## Non-audited: VM-neutral byte/XOR facts

The OpenVM-shaped byte-check facts are replaced by one layout-generic descriptor,
`ByteXorSpec` — a `(op, o₁, o₂, r)` `decode`/`encode` reordering plus `xorOp`/`pairOp` acceptance
semantics and byte bound. OpenVM decodes `[x, y, z, op]`, SP1 decodes `[op, a, b, c]`; both prove
the descriptor by construction. The byte-check/XOR passes were migrated to consume it through shared
builders (`mkByteCheck`/`mkBytePair`) and a `byteXorSpec_decode_iff` lemma:

- **All eight** consuming passes migrated: `XorEqExtract`, `DigitFold`, `RedundantByteDrop`,
  `ByteCheckPack`, `SplitBytePair`, `TupleRange`, `BusPairCancel` (byte-check emission), and
  `SeqzCollapse` (the OpenVM `sltu x, 1` gadget's range-check bus, threaded through its
  `clusterBus`).
- **All** OpenVM-specific byte/XOR facts removed: `xorZeroEq`, `xorComplEq`, `xorZeroCheck`,
  `xorComplCheck`, `byteCheck`, `bytePairBus`, plus the `byteCheck1`/`byteCheck2` builders. Only the
  VM-neutral `byteXorSpec` remains.

On OpenVM the emitted/recognized messages are byte-identical to before, so effectiveness is
unchanged; the migration only widens where the passes apply. Full build green, no warnings.

https://claude.ai/code/session_01EEGcJ1ZTCxuWPaeEZj9zKw

### DIFF
--- a/ApcOptimizer/MemoryBus.lean
+++ b/ApcOptimizer/MemoryBus.lean
@@ -2,18 +2,17 @@ import ApcOptimizer.Spec
 
 set_option autoImplicit false
 
-/-! Helper functions to encode the semantics of memory buses, to be used to implement
-    `BusSemantics::admissible`. Memory buses are stateful buses that follow the pattern
-    that bus interactions come in pairs like:
-    `send(memBusId, (<address 1>, <timestamp>, <value>))`
-    `receive(memoryBusId, (<address 2>, <previous timestamp>, <previous value>))`
-    such that if `<address 1> = <address 2>`:
-    - `<timestamp> = <previous timestamp>`
-    - `<value> = <previous value>`
+/-! Helper definitions to encode the semantics of memory buses, used to implement
+    `BusSemantics.admissible`. Memory buses are stateful: bus interactions come in
+    `getPrevious`/`setNew` pairs — a `setNew` commits a cell's value, and the next access's
+    `getPrevious` to the same cell reads it back. So a `setNew` and the next same-address
+    `getPrevious`, with no intervening same-address interaction, carry equal payloads (address,
+    timestamp and value). Both memory reads and memory writes are such a `getPrevious`/`setNew`
+    pair (a read additionally constrains the two values to agree).
 
-    Using `admissibleMemoryBus` just *asserts* that this discipline is followed. In practice,
-    it could be enforced (e.g. by an argument like [1]), or it could be not enforced, but guaranteed
-    that the prover *can* satisfy the property (without sacrificing completeness).
+    `admissibleMemoryBus` just *asserts* that this discipline holds. In practice it could be
+    enforced (e.g. by an argument like [1]), or not enforced but guaranteed that the prover *can*
+    satisfy it (without sacrificing completeness).
 
     Note that this assumes that *bus interactions are ordered by time*!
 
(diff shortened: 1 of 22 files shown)


ApcOptimizer/MemoryBus.lean
```diff
@@ -27,20 +27,40 @@ structure MemoryBusShape where
...
+      false`). -/
+  writePos : Bool := true
+
+/-- The multiplicity of a *write* on this bus: `1` if `writePos`, else `-1`. A read has the
+    negation. -/
+def MemoryBusShape.writeMult (shape : MemoryBusShape) : ZMod p :=
+  if shape.writePos then 1 else -1
+
+/-- The write multiplicity is nonzero whenever `1 ≠ 0` (it is `±1`). -/
+theorem MemoryBusShape.writeMult_ne_zero (shape : MemoryBusShape) (hp1 : (1 : ZMod p) ≠ 0) :
```
georgwiese: Theorems should not be in the this file. Pls move it into the Implementation directory. Same for the one below (fixing a TODO).

ApcOptimizer/MemoryBus.lean
```diff
@@ -27,20 +27,40 @@ structure MemoryBusShape where
   /-- Payload positions forming the access key (e.g. OpenVM's two-limb address).
       Can be empty for a single global cell. -/
   addressFields : List Nat
+  /-- Whether a *write* (the producer of a cell's value, read back by the next access) is sent with
```
georgwiese: In this entire file, pls don't speak of read and write, but call it getPrevious and setNew instead. Both reads and writes are a pair of getPrevious and setNew (read constraints that the values are the same), so speaking of read and write in this context is confusing.

ApcOptimizer/MemoryBus.lean
```diff
@@ -27,20 +27,40 @@ structure MemoryBusShape where
...
+      multiplicity `1`. OpenVM writes with `1` and reads with `-1` (`writePos = true`, the default);
+      SP1 sends the previous record and receives the new one, so it writes with `-1` (`writePos =
+      false`). -/
+  writePos : Bool := true
```
georgwiese: This name is confusing. Maybe this could be `direction`, expecting an enum with variants `ReceiveThenSend` and `SendThenReceive`? I also don't think this should have a default argument (change the OpenVM semantics to be explicit about it).

ApcOptimizer/MemoryBus.lean
georgwiese: Done in ffc1ed1: both `admissibleMemoryBus.consecutive` and `MemoryBusShape.setNewMult_ne_zero` moved to `ApcOptimizer/Implementation/MemoryBusDrop.lean`, resolving the TODO. `MemoryBus.lean` is now definitions-only.

---

ApcOptimizer/MemoryBus.lean
georgwiese: Done in ffc1ed1: the file now speaks of `getPrevious`/`setNew` throughout instead of read/write, with the docstring noting that both reads and writes are a getPrevious/setNew pair (a read additionally constrains the two values to agree).

---

ApcOptimizer/MemoryBus.lean
georgwiese: Done in ffc1ed1: replaced `writePos : Bool := true` with `direction : MemoryBusDirection`, an enum with variants `sendThenReceive` / `receiveThenSend` and no default. `setNewMult` is derived from it, and both the OpenVM and SP1 semantics now name their direction explicitly (`.sendThenReceive` / `.receiveThenSend`).

---

PR #141, opened 2026-07-16
Decide p.Prime once per run via PrimeWitness

pacheco:
## What

Prime-gated passes each ran `decide (Nat.Prime p)` — a trial division (~√p steps on the field modulus) — once per invocation and again on every cleanup iteration, dozens of times per optimizer run. `RedundantByteDrop` decided it per byte-check operand (super-linear).

This adds `PrimeWitness p` (the `Bool` result of the decision plus its soundness proof) to `Basic.lean`, computed **once per run** via `PrimeWitness.of`, and threads it to every prime-gated pass so each branches on an O(1) `Bool` read instead of re-deciding.

## Correctness / effectiveness

Semantically identity: `pw.isPrime ≡ decide (Nat.Prime p)` and the branches are unchanged, so **outputs are unchanged** (effectiveness-neutral). Verified on a sample against main — every case produces identical `(vars, constraints, bus)`. `lake build` and `Scripts/check-proof-integrity.sh` pass (the correctness theorems still depend only on `propext, Classical.choice, Quot.sound`); no change to the audited spec or the benchmark. The soundness a pass gets from `pw.correct` is the same `Nat.Prime p` it previously obtained from `of_decide_eq_true`.

## Runtime

Removes the repeated trial divisions. Spot-check vs main on small cases (where the fixed per-invocation cost is proportionally largest): apc_019 319→294 ms, apc_005 1056→968 ms, apc_003 248→210 ms — roughly −8..−20% on the small blocks; the big blocks (dominated by real work) are ~flat. The **Runtime Bench** CI will confirm the full corpus here.

## Note

Rebased onto current main (post-#136); the three files #136 rewrote (FlagFold/DomainFold/BusPairCancel) were resolved by keeping #136's bodies and adding only the prime gate — the two changes are orthogonal.

### DIFF
--- a/ApcOptimizer/Implementation/Optimizer.lean
+++ b/ApcOptimizer/Implementation/Optimizer.lean
@@ -83,29 +83,32 @@ def preludePasses : List (String × VerifiedPassW p) :=
     correctness proof follows automatically from the pass's own `PassCorrect`, and the profiler
     picks up the new label for free (it consumes this same list). -/
 def cleanupPasses : List (String × VerifiedPassW p) :=
-  [ ("zeroWidthRange", ZeroWidthRange.zeroWidthRangePass.guardDegree),
+  -- One primality decision per optimizer run, threaded to every prime-gated pass below (they read
+  -- the `Bool` in O(1) instead of re-running `decide (Nat.Prime p)` per invocation per iteration).
+  let pw := PrimeWitness.of p
+  [ ("zeroWidthRange", (ZeroWidthRange.zeroWidthRangePass pw).guardDegree),
     ("xorEqExtract", XorEqExtract.xorEqExtractPass.guardDegree),
-    ("carryBranch", carryBranchPass.guardDegree),
+    ("carryBranch", (carryBranchPass pw).guardDegree),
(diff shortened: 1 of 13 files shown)


pacheco:
Rebased onto current main (`048bcaf` = #144, which follows #140). The one conflict was `Optimizer.lean`'s pass lists: #140 had moved `tupleRange` out of `cleanupPasses` into the coda, and #144 unwrapped it there (`tupleRangePass.guardDegree`, no `iterateToFixpoint`) — resolved by keeping this PR's `pw` threading on `busPairCancel`/`busPairCancelLate`/`redundantByteDrop` and dropping the stale cleanup-cycle `tupleRange` line; the coda keeps the unwrapped entry (tupleRange has no prime gate, so no `pw` needed there). Everything else auto-merged.

Re-validated after the rebase: `lake build` clean (zero warnings), `Scripts/check-proof-integrity.sh` green (correctness theorems still depend only on `{propext, Classical.choice, Quot.sound}`), and `opt-export` output byte-identical to main on wasm-eth `apc_096`/`apc_058` and openvm-eth `apc_001`/`apc_034`.

PR #146, opened 2026-07-16
Step 11: IndexedState substrate (stacked on #145)

pacheco:
**Stacked on #145** (FactStore) — base is `factstore`, not `main`. Review/merge #145 first; I'll retarget to `main` once it lands.

## What

First increment of runtime-ideas **step 11** (the dirtiness / worklist architecture), building on the FactStore substrate from #145. Adds the verified data structures a worklist needs; **nothing consumes them yet**, so behavior is byte-identical.

## Why this shape

A measurement (temporary instrumentation, not committed) over the cleanup fixpoint:

| case | invocations | no-ops | global-version-skippable |
|---|---:|---:|---:|
| apc_019 | 104 | 67 (64%) | **0 (0%)** |
| apc_007 | 104 | 64 (61%) | **0 (0%)** |
| apc_028 | 130 | 80 (61%) | **0 (0%)** |

~61% of pass invocations are **no-op full-system scans** — the runtime prize. But the cheap "skip a pass that no-op'd at an unchanged global version" idea catches **0%**: the fixpoint only *retains* cycles that change something, so the global version bumps every cycle. Capturing the 61% needs **fine-grained, edit-stable dirtiness** — a powdr-style per-item worklist where a substitution over variables V dirties only the items mentioning V. That needs a representation that survives edits without reindexing.

## This PR: the substrate (`OptimizerPasses/IndexedState.lean`)

- `IndexedState p`: constraints and bus interactions as `Array (Option ..)` with **stable positions + tombstones** (a drop nils its slot, so occurrence positions never shift), a variable→positions **occurrence index**, and incremental **(var/constraint/bus/maxDegree) counters**.
- `ofSystem` / `toSystem` with the proven **round-trip `toSystem (ofSystem cs) = cs`** — the anchor a worklist uses to mutate the indexed state and materialize a genuine `ConstraintSystem` at the boundary for the existing `PassCorrect` machinery.
- `dropConstraintAt` / `dropBusAt` tombstone ops + their `toSystem` characterization.

Occurrence index and counters carry no threaded invariant (re-validated at use — the `CoveredIndex` discipline); the round-trip is the proven part. Carried in `FactStore.indexed : Option (IndexedState p)` (`none` until a consumer populates it).

## Next increments (not in this PR)

Wire a consumer: route the substitution-heavy core (gauss + domainBatch deductions + trivial cleanup) through a dirty-item worklist over the indexed state, saturating before materializing at the boundary — then extend to the hot scan passes. That is where the ~61% no-op scans get eliminated.

## Integrity

Build green, no warnings; audited surface and `Basic.lean` untouched; three correctness axioms only; `opt-export` byte-identical to main.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/FactStore.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/FactStore.lean
@@ -1,4 +1,5 @@
 import ApcOptimizer.Implementation.OptimizerPasses.FactPass
+import ApcOptimizer.Implementation.OptimizerPasses.IndexedState
 
 set_option autoImplicit false
 
@@ -80,9 +81,13 @@ def RangeCache.get (C : RangeCache p) (b : Nat) :
 structure FactStore (p : ℕ) where
   /-- Materialized `[0, bound)` range domains, shared across `domainBatch` invocations. -/
   rangeCache : RangeCache p
+  /-- Step-11 substrate: a stable-position, tombstoned, occurrence-indexed view of the system with a
+      proven round-trip to `ConstraintSystem` (`IndexedState`). `none` until a worklist increment
+      populates it; carried here so future passes can thread it. Not yet consumed. -/
+  indexed : Option (IndexedState p)
 
 /-- The empty store: no cached facts. -/
-def FactStore.empty : FactStore p := { rangeCache := RangeCache.empty }
+def FactStore.empty : FactStore p := { rangeCache := RangeCache.empty, indexed := none }
 
 instance : Inhabited (FactStore p) := ⟨FactStore.empty⟩
 
--- a/ApcOptimizer/Implementation/OptimizerPasses/IndexedState.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/IndexedState.lean
@@ -1,0 +1,128 @@
+import ApcOptimizer.Implementation.OptimizerPasses.FactPass
+
+set_option autoImplicit false
+
+/-! # Indexed cleanup state (step-11 substrate)
+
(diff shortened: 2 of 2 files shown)


pacheco:
**Status: backburner (do not merge yet).**

Measured the step-11 opportunity before building a consumer:
- ~61% of cleanup-pass invocations are no-op full-system scans, but a cheap global-version skip catches **0%** (the fixpoint retains only changing cycles).
- No-op invocations cost **14–22% of runtime** (apc_007 14%, apc_028 22%) — real, but spread thinly across ~8 passes (flagFold 429ms, busUnify 401, carryBranch 188, domainBatch 184, rootPairUnify 165…), no single pass dominant.
- Substitution locality: `gauss` changes ~71–80% of constraints per invocation (batch substitution isn't local → incremental substitution saves only the ~20% unaffected); `domainBatch` is local (~25%) but its `substF` is a small fraction of the pass.

**Conclusion:** this threaded `IndexedState` only pays via a broad *all-pass* rewrite (every mutating pass maintains the shared indices; every hot pass consults/skips via per-item dirtiness) — no cheap vertical slice captures a meaningful chunk. And much of the 14–22% is actually **local O(V²)/O(n²) inefficiencies inside individual passes** (flagFold's `eraseDups`, rootPairUnify's `seen`-join) that are fixable per-pass *without* this substrate (the doc's "step 6"), at far higher ROI.

So the foundation here is sound and byte-identical, but over-engineered for where the runtime actually is. Parking it: the shippable runtime win is the FactStore + range cache in #145. Revisit this branch only if/when the broad end-state rewrite is worth undertaking.

PR #147, opened 2026-07-16
Byte-boundary range re-split: ComputationMethod.floorDiv/floorMod + RangeResplit pass (entries 92–93)

georgwiese:
Re-encodes two-limb range decompositions at the byte boundary so the byte halves pack two-per-interaction. **Bus interactions drop on all three corpora — openvm-eth +0.131×, wasm-eth +0.073×, keccak 1,752 → 1,688 (46 below powdr's 1,734, previously documented as a floor) — with variables and constraints byte-neutral by design** (see "Why variables don't move" below).

The PR contains six commits:

1. **Entry 92 (docs only)**: a fresh-eyes structural pass that measured floors for every mass class and identified this transformation as the one big lever — blocked only by the `ComputationMethod` grammar.
2. **Spec extension (audited surface, user-approved)**: `ComputationMethod` gains `floorDiv (e) (d)` / `floorMod (e) (d)` — integer digit extractors on an expression's canonical value (~12 lines in `Spec.lean`, plus eval/vars/`eval_congr` cases and JSON export encodings). ⚠️ **powdr's witgen needs the twin extension** before consuming exported derivations.
3. **`RangeResplit`** (new coda pass) + a new proven fact `varRangeBusInv_sound` (on a variable-range bus, `breaksInvariant` ignores the payload — lets a pass *add* a check that inherits an existing check's multiplicity).
4. **`ByteCheckPack` generalization**: the single-byte recognizer accepts the bare `[e, 8]` variable-range form, and the packer picks any target bus carrying the pair facts (preferring the sources' own bus, so existing behavior is unchanged).
5. **Coda re-order**: `rangeResplit` runs after `tupleRange`, fixing a +1-bus parity effect the first CI run surfaced on wasm `apc_036`/`apc_063` (details below).
6. **Readable fresh-limb names**: `resplit_byte_<id>` / `resplit_high_<id>`.

## Worked example (`openvm-eth/apc_010`, register 44)

Register 44 is written inside the block. After the interior accesses telescope, its boundary pair remains: a receive carrying the cell's state *before* the block (dead `prev_data`, plus `prev_ts` — the last time anything touched this register before the block), and a send writing the final value at the last in-block access time. To stop replay, the circuit asserts `prev_ts < t_acc`, where **`t_acc` is the time of this register's first in-block access** (two ticks after block entry here; all in-block times are fixed offsets of the entry timestamp). The assertion is encoded, as usual, by a 29-bit decomposition of the difference:

```text
D = d_lo + 2^17·d_hi          # two private witness columns
range_check(d_lo, 17 bits)
range_check(d_hi, 12 bits)
memory_receive(reg 44, prev_data, prev_ts = t_acc − 1 − D)
```

`D ∈ [0, 2^29)` makes `prev_ts = t_acc − 1 − D` exactly equivalent to `prev_ts < t_acc` (no field wrap: 2^29 ≪ p). **Before**, this costs two full variable-range interactions per boundary address, because neither limb fits a shared slot: bitwise op-0 slots take exactly 8 bits and eth's tuple slot-2 exactly 13, so the 17-bit and the 12-bit check each burn a whole interaction.

`d_lo`/`d_hi` occur nowhere else — the two checks and that one linear form. So the *split* of `D` is arbitrary: any limb pair encoding the same 29-bit value works, and the byte-aligned split is strictly better for packing. **After**, the pass introduces two derived columns (filled by witness generation, not constrained):

```text
b := D mod 256        # a byte           (exported as ComputationMethod.floorMod)
h := D div 256        # 21 bits          (exported as ComputationMethod.floorDiv)
```

rewrites the receive to use them, checks `h` solo, and — the point of the exercise — the byte `b` shares one bitwise interaction with another address's byte:

```text
byte_pair_check(b, b₂)        # b₂ = the fresh byte of another access; 2 obligations, 1 interaction
range_check(h, 21 bits)
memory_receive(reg 44, prev_data, prev_ts = t_acc − 1 − (b + 256·h))
```

Per address: 2 interactions → 1.5. On apc_010 all 39 lt-pairs qualify: 78 solo range checks become 39 checks of `h` plus 20 byte-pair interactions — **total 239 → 220** (powdr keeps 239). Same mechanism on keccak's 129 boundary addresses: 1,752 → 1,688. In the exported circuit the fresh columns are named `resplit_byte_<id>` / `resplit_high_<id>` (id = the low limb's powdr ID).

(If you read the raw export you'll see a bare constant instead of `t_acc`: Gauss keeps one timestamp column as the anchor — here the fifth instruction's entry time — and writes every other time as anchor + constant. Block entry is anchor − 12 and this access is at anchor − 10, so the slot literally reads `anchor − 11 − D`.)

**Why this is sound**: `(d_lo, d_hi)` and `(b, h)` are both bijective digit decompositions of the same interval `[0, 2^29)`, so the satisfying sets correspond exactly in both directions — no check is weakened. The identity `n mod 256 + 256·(n div 256) = n` makes the witness mappings work unconditionally; the range bounds bind exactly where the (shared) multiplicity is nonzero. Pairs whose low limb carries meaning beyond `D` are rejected: the rewrite must eliminate both limbs everywhere or the pass doesn't fire (e.g. the mem_ptr alignment gadget also checks `limb₀ / 4`, so those pairs correctly stay untouched).

## Why variables don't move (CI Δ +0.000×)

The re-split swaps two old limb variables for two fresh ones per pair — variable-count-neutral by construction. Two witnesses per address is a real floor: a 29-bit obligation cannot be certified by one variable (the largest range-check slot is 25 bits). The win is deliberately on the second-priority axis, per the effectiveness rules ("a pass that cuts bus interactions without regressing variables is still an improvement").

## Measured (CI sticky comment on `ccdba67` vs main)

| corpus | variables | bus interactions | constraints |
|---|---|---|---|
| openvm-eth | ±0 | 3.557×/2.812× → **3.688×/2.990×** (72/100 cases improve; powdr 3.480×/2.822×) | ±0 |
| wasm-eth | ±0 | 6.099×/2.886× → **6.172×/2.928×** | ±0 |
| keccak | ±0 | 7.570× → **7.857×** (powdr 7.648×) | ±0 |

Runtime +1% (eth), −0% (wasm). Build and `Scripts/check-proof-integrity.sh` green throughout ({propext, Classical.choice, Quot.sound} only).

## Follow-up landed in this PR: coda re-order

The first CI run showed wasm `apc_036`/`apc_063` at +1 bus: on wasm (tuple slot-2 = 4096) a 12-bit limb could previously ride a tuple with a byte partner; re-splitting *before* `tupleRange` dissolved the 12-bit obligation and stranded that partner on odd parities. `rangeResplit` now runs **after** `tupleRange`: pairs the tuple packer consumes are count-equal either way, leftover bare pairs still re-split, nothing is stranded. Order-invariant on eth (slot-2 is 8192-exact) and keccak (no tuple interactions); spot-checked on the pushed binary: apc_010 = 220, wasm apc_036 back to 1029, wasm apc_004 at parity (146). The refreshed CI comment on the latest commit is the corpus-wide confirmation.

https://claude.ai/code/session_01XbGq1C4PaYfStsLxJBHut3

### DIFF
--- a/ApcOptimizer/Implementation/BusFacts.lean
+++ b/ApcOptimizer/Implementation/BusFacts.lean
@@ -186,6 +186,15 @@ structure BusFacts (p : ℕ) (bs : BusSemantics p) where
       ∀ (x b mult : ZMod p),
         bs.violatesConstraint { busId := busId, multiplicity := mult, payload := [x, b] }
             = false ↔ (b.val ≤ 25 ∧ x.val < 2 ^ b.val)
+  /-- On a `varRangeBus` bus, whether a message breaks an invariant depends only on its
+      multiplicity, never on its payload. Lets a pass *add* a range check whose multiplicity
+      expression matches an existing check's: the new message inherits the old one's
+      (non-)breaking status. `trivial` discharges it vacuously (`varRangeBus` is `false`). -/
+  varRangeBusInv_sound :
+    ∀ (busId : Nat), varRangeBus busId = true →
+      ∀ (x b x' b' mult : ZMod p),
+        bs.breaksInvariant { busId := busId, multiplicity := mult, payload := [x, b] }
+          = bs.breaksInvariant { busId := busId, multiplicity := mult, payload := [x', b'] }
   /-- Tuple-range-checker style *stateless* bus with fixed sizes: the 2-ary message `[x, y]` is
       accepted **iff** `x.val < s1 ∧ y.val < s2`, and at multiplicity `1` it never breaks an
       invariant. Lets a pass pack a byte obligation (`s1 = 256`) and an exact-width range check
@@ -303,6 +312,7 @@ def BusFacts.trivial (bs : BusSemantics p) : BusFacts p bs where
   xorZeroCheck_sound := by intro _ h; exact absurd h (by simp)
   varRangeBus _ := false
   varRangeBus_sound := by intro _ h; exact absurd h (by simp)
+  varRangeBusInv_sound := by intro _ h; exact absurd h (by simp)
   tupleRangeBus _ := none
   tupleRangeBus_sound := by intro _ _ _ h; exact absurd h (by simp)
   zeroCell _ := none
--- a/ApcOptimizer/Implementation/JsonSerializer.lean
+++ b/ApcOptimizer/Implementation/JsonSerializer.lean
@@ -84,7 +84,10 @@ def serializeExpr (m : Std.HashMap Variable Nat) : Expression p → Json
     (derived `serde` on `enum ComputationMethod<T, E>` in
     `constraint-solver/src/constraint_system.rs`):
     `const c → {"Constant": c}`, `quotientOrZero num den → {"QuotientOrZero": [num, den]}`,
-    `ifEqZero cond thenM elseM → {"IfEqZero": [cond, thenM, elseM]}`. -/
+    `ifEqZero cond thenM elseM → {"IfEqZero": [cond, thenM, elseM]}`,
+    `floorDiv e d → {"FloorDiv": [e, d]}`, `floorMod e d → {"FloorMod": [e, d]}` (the two
+    integer digit extractors added for range re-splitting; powdr's witgen must interpret them as
+    `⌊e.val / d⌋` / `e.val % d` on canonical values). -/
 def serializeComputationMethod (m : Std.HashMap Variable Nat) :
     ComputationMethod p → Json
   | .const c => Json.mkObj [("Constant", constJson c)]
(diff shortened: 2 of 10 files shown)


georgwiese:
This PR exploits a bug in the spec: While OpenVM *can* be configured to support a 21-Bit range check, it does not by default and is not in practice. Therefore, this optimization does not work. #155 fixes the spec.

Still, very cool to see a novel optimization of which there are no similar examples in the Rust implementation and which take a few reasoning steps to arrive to. Impressive!

PR #149, opened 2026-07-16
Runtime: hoist ZMod ring ops in Expression.fold (byte-identical)

pacheco:
Two **universal, output-identical** primitive optimizations, found via a sampling profile (perf/LBR) of the optimizer on openvm-eth apc_005 and the keccak stress case. Unlike a per-pass fix, these target costs *every* pass pays.

## A. `Expression.fold`: derive the ZMod ring ops once per call
`fold` — and therefore `constValue?` and the 8 passes that call `fold` — reconstructed the `ZMod p` ring dictionary at **every** constant-fold node. The keccak profile showed `constValue? → foldAdd → …CommRing… → lean_alloc_object` as a top allocation source (ZMod dict reconstruction ~7.5% self, feeding the ~45% RC/alloc churn). `fold` now extracts `Add`/`Mul` **once** (`foldWith`, the `evalFast`/`evalWith` pattern already used in `DomainBatch`/`Reencode`) and threads them through the recursion. `fold`/`fold_eval`/`fold_vars` keep their signatures — no consumer changes — and are proved extensionally equal via `foldWith_eval`/`foldWith_vars`.

## B. `Variable` `Hashable`: id-first
Hash the powdr id first (a `Nat`), skipping the `String` hash when present. powdr ids are unique per variable (`next_free_id`), so this is a cheaper, well-distributed hash, consistent with `BEq` (equal variables share both fields). The profile showed `lean_string_hash` ~2% plus `instDecidableEqVariable_decEq` (String-first) driven partly by `HashSet` collisions.

## Correctness / scope
Byte-identical — same decisions, same output. No changes to any audited surface; no proof obligations altered (both are Implementation-only). `lake build` clean, no warnings. Local spot-check: openvm apc_001 117→18 constraints (6.5×), apc_005 1424/585/694, wasm apc_001 126→16 (7.875×) — all identical to `main`.

Draft — opening for CI's full runtime + effectiveness + proof-integrity run; the sticky comment gives the corpus-wide main-vs-branch runtime A/B.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/ConstantFold.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/ConstantFold.lean
@@ -23,76 +23,98 @@ first pass, and is re-run after substitution passes to propagate freshly-exposed
 
 variable {p : ℕ}
 
-/-- Smart addition: fold two constants and drop `+ 0` on either side. -/
-def Expression.foldAdd (a b : Expression p) : Expression p :=
+/-- Smart addition with the field `add` supplied by the caller (so the `ZMod p` ring dictionary is
+    derived once, not per node): fold two constants (via `add`) and drop `+ 0` on either side. -/
+def Expression.foldAddWith (add : ZMod p → ZMod p → ZMod p) (a b : Expression p) : Expression p :=
   match a, b with
-  | .const m, .const n => .const (m + n)
+  | .const m, .const n => .const (add m n)
   | .const m, b => if m = 0 then b else .add (.const m) b
   | a, .const n => if n = 0 then a else .add a (.const n)
   | a, b => .add a b
 
-/-- Smart multiplication: fold two constants, absorb `* 0`, drop `* 1`. -/
-def Expression.foldMul (a b : Expression p) : Expression p :=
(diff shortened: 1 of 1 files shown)


pacheco:
## CI verdict: not a win — closing

The quick-bench is single-run and ±7% noisy (main baseline swung 173.3s↔180.1s between two runs), but the **per-stage pattern is consistent across both runs**: `constValue?`/`fold`-heavy passes regress (rootPairUnify 1.23–1.27×, busPairCancel 1.29–1.32×, redundantByteDrop 1.33×, subsumedRange ~1.5×), while passes that don't call `fold` are neutral.

Root cause: the premise was wrong. The `ZMod p` ring dictionary is **not** reconstructed per `fold` call — it's threaded in as an instance-dictionary parameter down the call chain. So hoisting it saved nothing and *added* closure-indirection overhead on every `foldAddWith`/`foldMulWith`. The `evalWith` hoist pattern pays off in tight loops over many points (domain enumeration), not in `fold`.

Effectiveness stayed byte-identical (+0.000× everywhere), so the change was correct — just not beneficial. Closing rather than leaving a regression parked as a draft.

PR #150, opened 2026-07-17
Runtime: symbolic finite domains in domainBatch (defer range materialization)

pacheco:
## What

Represent a variable's finite domain **symbolically** — `FiniteDomain = explicit (List) | range (bound)` — and **stream** the box enumeration instead of materializing it.

- **Before:** a fact-bounded 16-bit slot materialized a 65536-element `(List.range b).map Nat.cast` at table-build time (shared per-bound via `RangeCache`); every box-size test rescanned those lists with O(n) `List.length`; the joint enumeration materialized the whole Cartesian product before scanning.
- **After:**
  - The table stores `FiniteDomain`; a `range` keeps only its `Nat` bound → table build materializes nothing, `RangeCache` deleted, every box-size test is an O(1) `size` read.
  - `boxFold` streams the Cartesian product with an early exit: a `Nat` loop (`rangeFoldFrom`) enumerates each `range` factor with **no element list built**, stopping the instant the consumer is done. The scan (`scanBox`) and redundancy check (`allBox`) use it, so a target that forces nothing over a 2^16 box aborts after a couple of points instead of materializing + scanning 65536.
  - `toList` / `matList` / `assignments` no longer appear on the runtime range path — `toList` is kept only as the proof specification.

## Correctness — byte-identical

`boxFold` / `scanBox` / `allBox` are proved equal to the eager `foldlStop … (assignments (matList doms))` in the original enumeration order (`boxFold_eq`, `scanBox_eq`, `allBox_eq`), and `(range b).toList = (List.range b).map Nat.cast` with `size = toList.length`. So every table entry, box-size test, redundancy decision, and forced value is unchanged. The three scan-soundness lemmas were decoupled from `DomainTable` (they take the materialized domains + their two properties); `scanBox_eq` / `matList_map_fst` bridge the lazy driver to them. Audited surface untouched; `lake build` clean, no warnings. Verified identical optimizer output on 24 openvm-eth/wasm-eth cases locally (incl. the range-heavy apc_019/apc_079).

## Perf

Removes the table-build range materialization, the repeated O(n) `List.length` box scans, and — via the early-exit `Nat`-loop enumeration — the full-product materialization for aborted scans. Deferring to CI for authoritative main-vs-branch corpus numbers (shared local machine).

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/DomainBatch.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/DomainBatch.lean
@@ -27,12 +27,44 @@ like `domainPropPass`); identity otherwise. -/
 
 variable {p : ℕ}
 
+/-! ## Symbolic finite domains
+
+A variable's finite domain is either an **explicit** list of field elements (the roots of a
+product-of-affine-factors constraint) or the **range** `[0, bound)` cast into `ZMod p` (a
+fact-bounded raw payload slot). A range keeps only its `Nat` bound: its `bound`-element list —
+`(List.range bound).map Nat.cast`, a `2^16`-element materialization for a 16-bit limb — is
+built **only** when a target is actually enumerated (`toList`, inside `assignments`), and its
+`size` is read in O(1) without materializing anything. Every table/box decision that used
+`List.length` on the old materialized list now reads `size`; `size_eq` proves them equal, so
+the pass is byte-identical to the eager version. -/
+
+/-- A finite domain, kept symbolically: an explicit element list, or the range `[0, bound)`. -/
+inductive FiniteDomain (p : ℕ) where
+  | explicit (values : List (ZMod p))
+  | range (bound : Nat)
+
+/-- The materialized element list of a finite domain (the range is cast lazily, only here). -/
+@[inline] def FiniteDomain.toList : FiniteDomain p → List (ZMod p)
+  | .explicit vs => vs
+  | .range b => (List.range b).map (Nat.cast : Nat → ZMod p)
+
+/-- The domain's cardinality — O(1) for a range (no materialization). -/
+@[inline] def FiniteDomain.size : FiniteDomain p → Nat
+  | .explicit vs => vs.length
+  | .range b => b
+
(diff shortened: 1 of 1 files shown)


pacheco:
Thanks — addressed the review by finishing the vertical slice rather than reintroducing `RangeCache`:

- Added `FiniteDomain.foldElts` with a `Nat` loop (`rangeFoldFrom`) for `.range` — no element list is built.
- Added `boxFold`, a lazy Cartesian-product fold that generates one assignment at a time and carries an early-stop state, used by both `allBox` (redundancy check) and `scanBox` (the box scan) in `constraintRedundant` and `forcedOver`.
- The runtime range path no longer calls `FiniteDomain.toList`, `DomainTable.matList`, or `assignments`; `toList` remains only as the proof specification.
- Proved the lazy fold equal to enumeration over `assignments (matList fdoms)` in the existing order (`boxFold_eq`, and `scanBox_eq` / `allBox_eq` on top), so the current soundness results are reused unchanged.

This eliminates the repeated-materialization concern entirely (nothing is materialized on the range path) and short-circuits aborted scans.

pacheco:
Both points addressed:

1. **Stale comments fixed** (d3a9c44, comment-only). The `FiniteDomain` section header, the `toList`/`matList` docstrings, and the `forcedOver` box-size comment now state that the runtime range path builds nothing — `boxFold` streams a range with a `Nat` loop, and `toList`/`matList` are proof specifications only.

2. **Nat→ZMod recast tradeoff** — agreed, this is the intended tradeoff (per-visit cast vs. the old cache's shared materialized values), and CI bears out that it's net-positive: `domainBatch` **3723ms → 2733ms (0.73×)** with **no per-case regression** across 200 cases, so full-box scans that recompute casts don't dominate in practice (the cast is a single sub-2³¹ `ZMod` construction, and most range boxes abort early). Effectiveness is byte-identical on all three corpuses. It stays a monitorable tradeoff — if a future workload shows many full-box scans over a shared bound, a per-invocation cast cache on the hot bound could be revisited, but the measurements don't call for it now.

PR #153, opened 2026-07-17
Runtime: remove the quadratic algebra cores in Gauss/normalize (output byte-identical)

pacheco:
## What

Removes the **asymptotic quadratic** behaviour in the affine / Gaussian-elimination algebra core. Short affine forms hide these costs, but a large enough block exercises them — e.g. the ~85 MB sha256 APC that currently takes hours — so the goal here is to remove the *complexity*, not just the hot constant.

Each disjoint change is its own commit, and **all are output byte-identical** (verified by the CI effectiveness check: exact per-case circuit sizes, identical to `main` on all three corpuses).

| commit | quadratic removed | how |
|---|---|---|
| build only the winning pivot | candidate `toExpr` construction, **O(#candidates·terms)** per constraint | `fastBest` scores lightweight `(variable, score)` descriptors and builds a solution `Expression` only for the `argmin` winner; proven equal to `(pm1PivotsOf ++ unitPivotsOf).argmin pivotScore` |
| linear like-term merge | `mergeTerms`/`addCoeff`, **O(terms²)** | `mergeTermsFast`: HashMap coefficient accumulation + first-occurrence order array, O(terms); proven `= mergeTerms`, installed via `@[csimp]` (no call-site churn) |
| linear pivot scoring | per-candidate `l.coeff`/`l.others` rescans, **O(terms²)** per constraint | one `(coeff, occurrence-count)` index (`coeffIdx`) built once per constraint, O(1) lookups → O(terms); proven equal to the coeff/others formulation |
| reverse-dependency index | per-pivot solution-map scan `σ.map.toList.filter (·.mentions x)`, **O(K²)** over K pivots | resolve a new pivot against only `x`'s reverse-dependency bucket (`HashSet` buckets, idempotent re-posting) instead of the whole map |

The **blind double sweep** (`constraints ++ constraints`) is deliberately left as-is: it is a 2× *constant* factor and is load-bearing — a single sweep *regresses* (it defers pivot discovery to later cleanup iterations over larger systems) — not an asymptotic quadratic.

## Scope & soundness

- Implementation only: `Gauss.lean` and `Normalize.lean`. No audited surface, benchmark, or pass schedule touched.
- Every pass remains a `VerifiedPass` with its machine-checked `PassCorrect`; the top-level correctness theorems depend only on the three standard axioms. `Scripts/check-proof-integrity.sh` passes — no `sorry`/`admit`/`native_decide`, no new axioms, no new `partial`.
- The first three commits prove the fast path *equals* its reference spec. The fourth adds `revDeps` as a **data-only** field: the pass stays sound via the same per-pivot entailment proof (a stale/incomplete bucket could only make the pass eliminate *fewer* variables, never emit a wrong circuit), and the index's completeness — hence full resolution, hence byte-identical output — is verified by the exact per-case size comparison, the "retain the entailment proof + strong output comparison in tests" option the task permits (as in the merged busPairCancel stable-array change).

## Effectiveness & runtime

- **Effectiveness: exact** — per-case circuit sizes identical to `main` on openvm-eth, wasm-eth, and keccak (+0.000× everywhere).
- On the current short-form corpus these are ~neutral (short affine forms don't trigger the O(terms²) cores), as expected; the payoff is asymptotic, for large blocks like the sha256 APC. The one visible signal is the `gauss` pass stage at ~0.87× (−13%, from the build-only-winner change). Aggregate runtime rows move within the ±7% single-run quick-bench noise.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/Gauss.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/Gauss.lean
@@ -155,7 +155,249 @@ def pivotScore (occ : Std.HashMap Variable Nat) (prot : Std.HashSet Variable)
   let base := (occ.getD xt.1 1 - 1) * (1 + xt.2.varCount)
   if prot.contains xt.1 && !xt.2.isVar then base + 1000000 else base
 
-/-! ## The elimination loop -/
+/-! ## Fast pivot selection (build only the winning candidate)
+
+`pm1PivotsOf`/`unitPivotsOf` build a full solution `Expression` (`toExpr`) for *every* solvable
+pivot of a constraint, then `argmin` keeps one and discards the rest — wasteful expression
+allocation, which dominates via reference-count churn on large systems. `fastBest` scans
+lightweight `(variable, score)` descriptors instead, computing each candidate's exact `pivotScore`
+from the linear form's coefficients and term counts *without materialising any solution*, then
+builds `toExpr` only for the winner. It is proven equal to the old `argmin` over
+`pm1PivotsOf ++ unitPivotsOf` (`fastBest_eq`), so `gaussLoop`'s decisions and output are unchanged.
+
+Remaining opportunity (not taken here): `pm1Desc`/`unitDesc` still build `l.others v` and rescan
+`l.coeff v` per candidate, so scoring is O(terms²) per constraint. On the measured corpora affine
+forms are short, so this is below the win; a single coefficient/others-count index pass would make
+scoring linear if longer forms ever dominate. -/
+
+/-- `argmin` commutes with a key-preserving map: when `g` carries the key (`kγ (g a) = kα a`), the
+    winner of the mapped list is the mapped winner. This lets us score cheap descriptors in place
+    of built candidates. -/
+theorem argmin_map_key {α γ : Type*} (g : α → γ) (kα : α → Nat) (kγ : γ → Nat)
+    (h : ∀ a, kγ (g a) = kα a) : ∀ l : List α, (l.map g).argmin kγ = (l.argmin kα).map g := by
+  intro l
+  induction l with
+  | nil => simp
+  | cons a t ih =>
+      rw [List.map_cons, List.argmin_cons, List.argmin_cons, ih]
+      cases t.argmin kα with
+      | none => simp
+      | some c =>
+          simp only [Option.map_some, h]
+          by_cases hlt : kα c < kα a <;> simp [hlt]
+
+theorem map_filterMap {α β γ : Type*} (f : α → Option β) (g : β → γ) (l : List α) :
+    (l.filterMap f).map g = l.filterMap (fun a => (f a).map g) := by
+  induction l with
+  | nil => simp
+  | cons a t ih =>
+      simp only [List.filterMap_cons]
+      cases f a with
+      | none => simpa using ih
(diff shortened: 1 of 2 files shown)


pacheco:
## Reading the bench

**Effectiveness is exact** — "Per-case circuit sizes are identical between main and this branch" on all three corpuses (+0.000× everywhere). This is a runtime-only, byte-identical change, as intended.

**The targeted win is confirmed at the pass level:** openvm-eth per-stage shows **gauss 1956 → 1700 ms (0.87×, −13%)**, with every other pass at ~1.00× (this PR only touches `Gauss.lean`). Every top-10 openvm case is faster (0.96–0.99×).

**The aggregate `runtime` total (openvm +5%) is single-run quick-bench noise, not a regression:**
- the per-case table shows all 10 slowest cases *faster* (sum ≈ −0.46 s), and
- the per-stage sum is net lower (gauss −256 ms, everything else within ±3%).

A +5% aggregate while every measured case and stage is flat-or-faster is inconsistent, i.e. it's the ~±7% variance of the whole-corpus single run, not real work added. The per-stage `gauss 0.87×` is the robust signal.

Happy to trigger the best-of-N `Runtime Bench` (`gh workflow run "Runtime Bench" -f pr=153`) for a lower-variance aggregate if useful.

pacheco:
## Expanded to remove the asymptotic quadratics (per the task), as separate commits

Following the discussion of the 5 h sha256 APC (85 MB), the goal here is to *actually remove* the algebra core's asymptotic quadratic behaviour (not just the hot allocation), since a large-enough block will exercise it. Each disjoint change is its own commit; all are output byte-identical.

1. **`ed5ca59` — build only the winning pivot.** Removes O(#candidates·terms) candidate-`toExpr` construction (build only the `argmin` winner).
2. **`0708209` — linear like-term merge (`mergeTermsFast`).** `mergeTerms`/`addCoeff` were O(terms²) (linear scan of the accumulated list per term); now a HashMap coefficient accumulation + first-occurrence order array, O(terms). `@[csimp]`-installed, proven `= mergeTerms`.
3. **`ecc82b6` — linear pivot scoring.** Scoring called `l.coeff v`/`l.others v` (O(terms)) per candidate → O(terms²)/constraint. Now one `(coeff, occurrence-count)` index (`coeffIdx`) built once per constraint, O(1) lookups → O(terms). Proven equal to the coeff/others formulation.
4. **`ac7b49a` — reverse-dependency index.** Per-pivot solution resolution scanned the whole solution map (`σ.map.toList.filter (·.mentions x)`) → O(K²) over K pivots. Now scans only `x`'s reverse-dependency bucket. `HashSet` buckets (idempotent re-posting; `List` buckets accumulated duplicates and blew up). Data-only field: the pass stays sound via the same per-pivot entailment proof; completeness/full-resolution is verified by the exact per-case size comparison in the effectiveness check.

The blind double-sweep is left as-is — it's a 2× constant factor and load-bearing (a single sweep *regresses*, since it defers pivot discovery to later cleanup iterations over larger systems), not an asymptotic quadratic.

On the current short-form corpus these are ~neutral (apc_037 byte-identical, ~19.4 s) — expected, since short affine forms don't exercise O(terms²); the win is asymptotic, for large blocks like the sha256 APC. Soundness proofs unchanged (three standard axioms); the effectiveness check is the full-corpus byte-identity gate.

PR #154, opened 2026-07-17
Runtime: de-quadratify seen-scans, dedup, and enumeration eval (effectiveness unchanged)

georgwiese:
## What this does

Pure **runtime** work targeting the optimizer's super-linear scaling: the *output is unchanged* on
every benchmark case, so effectiveness is identical — only wall-clock cost drops. Nothing in the
audited surface, `Basic.lean`, or the spec changed; correctness follows from each pass's own
`PassCorrect`. `lake build` green, no warnings, no `sorry`/`native_decide`.

## Motivation

The nightly powdr run showed the optimizer scaling ~quadratically — on openvm-eth, ~14× more
constraints costs ~100× more optimizer time. The super-linear work is a set of **O(n²) `seen`-scans
/ dedup** (each element checked against a growing history) plus a **quadratic finite-domain
enumeration eval** and per-invocation **whole-tree rebuilds**.

## Changes (each output-preserving; verified byte-identical vs `main` on all 100 openvm-eth cases + keccak)

**De-quadratified O(n²)→~O(n) via hash-bucketing.** Each of these passes picks a candidate then
re-checks it with a certificate, so the *search* carries no proof obligation; bucketing the history
into a `HashMap` keyed by a hash of the match key (equal keys share a bucket, collisions resolved by
the retained exact check, insertion order preserved) returns the identical picked element:
- `rootPairUnify` (`rpLoop`) and `flagUnify` / `flagFold` part A (`fuLoop`/`fxLoop`) — the two-root
  and scaled-check candidate `seen`-scans.
- **`dedup` bus side** (`dedupStateless`): the `bi ∈ seen` membership scan → per-bucket scan.
  `dedupStatelessFast_eq` proves (a HashMap↔list-agreement induction) it returns the identical list,
  and `dedupFast_eq` reuses `dedup_correct` verbatim. **keccak: dedup 16.6 s → 7.0 s.**

**Finite-domain enumeration** (`groupSurvivorsE`, shared by `reencode` + `domainFold`): compile the
covered constraints once per target to positional `IExpr` with the ring ops hoisted out of the
instances (the `domainBatch` `compiledSurv` treatment) instead of `evalFast (envOf a)` at every box
point; `groupSurvivorsE_eq` proves the identical survivor list. **eth apc_100: reencode 7417 → 5940
ms, domainFold 2652 → 2227 ms.**

**Recomputation / rebuild hoists (proof-transparent `let`s):**
- `domainFold` (`foldStep`): compute `foldOut` once per accepted fold (was twice).
- `domainBatch` (`forcedOver`): hoist the recomputed domain-key list.
- `flagFold` part B (`boxTautoDrop`): build the per-variable domain cache in one linear pass instead
  of an O(N²) `eraseDups` over the whole occurrence list.

## Results

- **Effectiveness unchanged** — output byte-identical vs `main` on all 100 openvm-eth cases and on
  the keccak stress case (`2021 / 186 / 1752`), via `apc-optimizer run`.
- **openvm-eth optimizer time ≈ −5%** aggregate over all 100 cases, no per-case regressions; the
  slowest cases benefit most (the enumeration + `seen`-scan fixes: large cases ≈ −5 to −8%).
- **keccak ≈ −7%** (best-of-2 runs; single-run variance ±3%), led by the dedup-bus (−9.6 s) and the
  `seen`-scan bucketing.
- The `seen`-scan and dedup fixes are **asymptotic** (O(n²) → ~O(n)), so they help larger
  (SHA-scale) circuits disproportionately more than the keccak/eth percentages alone show.

## Not included (honest frontier)

- The enumeration trio (`domainBatch`/`domainFold`/`reencode`) dominates the keccak profile, but on
  keccak (>8192 constraints) it runs its inverted-index paths, already near the engineered floor; the
  residual is the index gather/rebuild, needing a position-stable `reencodeOut` (log entry 86).
- `busPairCancel` (O(B²) shielded-prefix rebuild, just optimized in #151) and `busUnify` (O(B²)
  forward consumer scan) are **positional/consecutive** scans, not value-bucketable.
- `gauss`'s residual O(V²) resolved-map maintenance (`σ.map.toList` per pivot) needs a proof-carrying
  reverse index with a completeness invariant through resolution. (A substF-rebuild-skip for
  solved-var-disjoint constraints was implemented and measured **neutral** on keccak, so it was
  dropped.)

https://claude.ai/code/session_01NxTz3FwUa6oTALyWZtNsQV

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/Dedup.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/Dedup.lean
@@ -118,6 +118,74 @@ theorem dedupStateless_evalFilter (bs : BusSemantics p) (env : Variable → ZMod
         List.map_cons,
         List.filter_cons_of_neg (by have hf : bs.isStateful (b.eval env).busId = false := (by simpa using h1); simp [hf]), ih]
 
+/-! ## Hash-bucketed stateless dedup
+
+`dedupStateless`' membership test `bi ∈ seen` scans the whole kept-so-far list — quadratic in the
+number of stateless interactions. Bucketing `seen` by a structural hash of the interaction turns
+each test into a one-bucket scan; `dedupStatelessFast_eq` proves the bucketed version returns the
+identical list, so the pass output — and the whole `dedup_correct` proof — are unchanged. -/
+
+/-- A structural hash of an expression (for bucketing). -/
+def Expression.bHash : Expression p → UInt64
+  | .const n => mixHash 11 (hash n.val)
+  | .var x => mixHash 13 (hash x)
+  | .add a b => mixHash 17 (mixHash a.bHash b.bHash)
+  | .mul a b => mixHash 19 (mixHash a.bHash b.bHash)
+
+/-- A structural hash of an interaction (bus id, multiplicity, payload). -/
+def BusInteraction.bHash (bi : BusInteraction (Expression p)) : UInt64 :=
+  mixHash (hash bi.busId)
+    (mixHash bi.multiplicity.bHash (bi.payload.foldl (fun h e => mixHash h e.bHash) 7))
+
+/-- `dedupStateless` with the `seen` set bucketed by `BusInteraction.bHash`: each membership test
+    scans only the matching bucket. -/
+def dedupStatelessFast (bs : BusSemantics p)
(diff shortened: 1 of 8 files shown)


georgwiese:
Modest wins, and promises to help with SHA (which takes 5h currently). Will have to double check, but seems plausible.

PR #155, opened 2026-07-17
Fix variable range checker spec

georgwiese:
The up to 25-Bit range check was to generous. While OpenVM *can* be configured that way, it is not by default and in practice. #147 exploited this!


Schaeff:
Would it be of any value to define a constant for this?

PR #161, opened 2026-07-18
SP1: drain dead-byte clusters + degenerate range checks (vars 3.672×→3.745×, bus 2.488×→2.553×)

georgwiese:
Three proven effectiveness wins on the SP1 benchmarks, each a **generalization of an existing mechanism** (no new bespoke passes where a generalization sufficed). No audited surface touched; no `sorry`/`axiom`/`native_decide`; proof integrity green (`{propext, Classical.choice, Quot.sound}`). **OpenVM keccak byte-identical throughout** (2021 v / 186 c / 1752 bus).

## Changes

**1. Dead-byte OR/AND clusters** (`BusFacts`, `Sp1Facts`, `OpenVmFacts`, `XorEqExtract`, `ScaledZero`)
The SP1 `lbu; xor; sb` dead bytes were handled for XOR but not OR/AND, and their operands are two-term scaled forms `8323072·(b_low − higher)` that `scaledZero` (single scaled variable) could not force.
- `xorEqExtract` generalized XOR→OR/AND: `ByteXorSpec` gains optional `orOp`/`andOp` selectors (SP1 `1`/`0`; OpenVM `none`) + a new `byteBoolSound` fact. The OR arm fires only when the surviving operand is a **constant** (a wider target displaces the byte apc's memory cancellation keys on — measured regression otherwise).
- `scaledZero` gains the two-term slot `k·v − k·w` (`two_term_zero` + `pair2Seeds`): a byte slot of that shape with `k ≥ B2` and no wrap forces `v = w`; Gauss merges the limbs, the emptied operand reaches `0`, and the OR arm folds `result = 0`.

**2. Width-0 op-6 → equality** (`RangeForceZero.lean`)
SP1's degenerate width-0 range check `[6, L, 0, 0]` (`L < 1`, i.e. `L = 0`) — often a decomposition equality `higher_limb = result₀ + 256·result₁ + …` — is seeded as `L = 0` via the layout-agnostic `rangeCheckAt` fact; Gauss consumes it, `tautoBus` drops the now-constant check. The `rangeCheckAt` generalization of `ZeroWidthRange`'s width-0 arm.

**3. Width-1 op-6 → booleanity** (`RangeBool.lean`)
Width-1 check `[6, x, 1, 0]` on a bare variable → `x·(x−1) = 0` + drop (two-step `PassCorrect`, prime-gated, degree-2), the `rangeCheckAt` half of `ZeroWidthRange`'s booleanity arm. Reuses `ZeroWidthRange.boolC` / `val_lt_two_iff`.

OpenVM declares no `orOp`/`andOp`/`rangeCheckAt`, so all three are no-ops there.

## Impact (`benchmark.py --vm sp1`, 100 rsp cases)

| measure | before | after | powdr |
|---|---|---|---|
| variables | 3.672× | **3.745×** | 3.980× |
| bus interactions | 2.488× | **2.553×** | 2.822× |

Aggregate var gap vs powdr **880 → 658**, bus gap **976 → 765** (~25 % each), **0 regressions**. SP1 keccak also improved (3586→3564 v, 2489→2409 bus).

## Remaining gap (documented in `docs/ideas.md`)

The k256 residual is two architectural levers, each investigated and found to need more than an incremental pass: (a) **identity-result elimination** — the extraction is landed and proven, but a bare-variable target triggers a range-check re-packing / `reencode` explosion (memory is untouched); (b) **memory telescoping** of the negative-coefficient data slots, which are 16-bit only modulo a same-address relation powdr's global memory argument establishes but apc's local pair-cancellation cannot.

---
_Generated by [Claude Code](https://claude.ai/code/session_0172asXbkaAHzFPCoVAPt4R4)_

### DIFF
--- a/ApcOptimizer/Implementation/BusFacts.lean
+++ b/ApcOptimizer/Implementation/BusFacts.lean
@@ -45,6 +45,13 @@ structure ByteXorSpec (p : ℕ) where
   xorOp : ZMod p
   /-- Op-selector value denoting the pair range-check (`op₁, op₂` bytes, `result = 0`). -/
   pairOp : ZMod p
+  /-- Op-selector value denoting the bitwise-OR relation `result = op₁ | op₂` (`op₁, op₂` bytes), if
+      the bus has such an op. `none` when the VM's bitwise bus has no OR op (e.g. OpenVM's, which is
+      XOR + range only). Soundness lives in `BusFacts.byteBoolSound`. -/
+  orOp : Option (ZMod p) := none
+  /-- Op-selector value denoting the bitwise-AND relation `result = op₁ & op₂` (`op₁, op₂` bytes), if
+      the bus has such an op. `none` when the VM's bitwise bus has no AND op. -/
+  andOp : Option (ZMod p) := none
   /-- Reorder a physical payload into logical `(op, operand₁, operand₂, result)`. -/
   decode : {α : Type} → List α → Option (α × α × α × α)
   /-- Build a physical payload from logical `(op, operand₁, operand₂, result)` — the inverse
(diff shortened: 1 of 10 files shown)


georgwiese:
Cool to see a slight improvement on OpenVM too!

PR #163, opened 2026-07-18
SP1: interval forcing + basis justification + OR-identity checks (vars 3.78×→3.92×, bus 2.55×→2.70×)

georgwiese:
Three generalizations of existing mechanisms, found by fresh per-case export diffing against powdr (not the prior gap attributions, two of which turned out wrong). Cumulative on SP1 rsp: **variables 3.784× → 3.922×** (aggregate var gap vs powdr 442 → 55, per-case W/L/T 1/51/48 → 15/42/43), **bus 2.553× → 2.703×** (bus gap 765 → 319), **0 per-case regressions** across all iterations. Bonuses: **sp1/keccak vars 3.96× → 5.16×** (now ahead of powdr's 4.81×) and bus −236; wasm-eth improves on all axes; OpenVM keccak −4 bus; openvm-eth byte-identical.

### 1. `intervalForce` (entry 101) — replaces `ScaledZero`

For every accepted bus slot with a `slotBound B` whose expression is affine in range-bounded variables (and every algebraic constraint, as a bound-1 slot), take signed minimal-magnitude coefficient representatives; if the integer window spanned by the bounds survives reduction mod `p` only at residues `[0, B)`, the slot value is an **integer** in `[0, B)`, and coefficient pairs `+g·v/−g·w` (window conditions `B ≤ g + Rlo`, `Rhi < g`) force `v = w`; single terms force `v = 0`. Seeds go in as constraints; Gauss cascades. SP1's op-6 checks `r₀ + 256·r₁ − 256·h < 2⁸` force `h = r₁` — the `higher_limb`/`c_low_bytes` clusters powdr substitutes away. ScaledZero's two arms are the 1- and 2-term instances, so its file is deleted.

### 2. `basisJustified` (entry 102) — a new arm of `byteJustifiedW`

The shift-result memory chains would not telescope because the dropped send's 16-bit obligation isn't affine in bounded *variables*: the slot is `16384·F + h₀` where `F = r₂ + 256·r₃ − 4·h₁` is itself op-6-checked `< 4`. The new arm reduces the target by positive integer multiples of *range-checked slot forms* (from an untrusted per-invocation position index, re-checked live at use), then finishes with the plain natural bound — one nat-level induction, no wraparound case analysis. Unlocks the srl register chains (−154 bus over 10 rsp cases) and fired on wasm-eth (apc_006 −128 bus) and OpenVM keccak (−4) too.

### 3. OR-identity byte checks (entry 103) — two recognizer arms + a coda reorder

After the entry-100 rename, SP1's k256 blocks hold piles of degenerate `[or, x, x, 0]` interactions (= "x is a byte"; 48 on apc_024 alone). `svCheck?` and `byteCheckOperands?` gain the OR-identity arms (sound via `byteBoolSound` lifted to symbolic payloads), and `identitySubst` moves before the coda's dedup/drop/pack stages, so the degenerate checks dedup, drop when justified, and pack pairwise into op-3 (−152 bus over 13 rsp cases). OpenVM declares no `orOp`, so it is untouched by construction.

### Runtime

Local serial A/B on the heaviest rsp cases is flat-to-better (apc_030 22.6 s → 21.9 s; `busPairCancel` itself 1791 → 913 ms — fewer surviving pairs to rescan; `intervalForce` ≈ 1 % of heavy-case profile). The CI matrix runtime row (parallel, report-only) shows openvm-eth +4 %, sp1/keccak −7 %, rsp/wasm ≈ +15 % on the iteration-2 run — if that persists on the final run, note the row's contention caveat; the per-pass breakdown shows no new hot pass.

### Docs

`docs/log.md` entries 101–103; `docs/ideas.md` fully rewritten (landed/obsolete items dropped, fresh priorities: quadratic root domains → SubsumedCheck basis arm → op-3 packing parity → variable tail → wasm range repack).

No audited files touched. Proof integrity green ({propext, Classical.choice, Quot.sound}); no `sorry`/`axiom`/`native_decide`.

https://claude.ai/code/session_0156XuePg7df6p6n7dH4zwak


georgwiese:
Looks great!

PR #166, opened 2026-07-20
Runtime: dense VarId internal representation for the whole optimizer pipeline (encode once, ID-native passes with bundled proofs, decode once)

pacheco:
Migrates the optimizer's internal representation from the original `Variable`-keyed form to a dense `VarId` (registry-indexed) form, end to end: the constraint system is **encoded once** at optimizer entry, all 42 scheduled passes (1 prelude, 29 cleanup, 12 coda) run **ID-native** with their own `DensePassCorrect` correctness proofs, and the result is **decoded once** at output. The audited `Variable`-based specification (`Spec.lean`, the VM semantics, `MemoryBus.lean`, `ApcOptimizer/Optimizer.lean`, `Basic.lean`) is untouched; correctness is lifted to it at exactly the pipeline edges via the `Bridge` round-trip.

What this PR contains, beyond the representation port itself:

- **Native proofs throughout** — every pass bundles a proof over `VarId → ZMod p` environments; the interim decode-commutation proof strategy and its scaffolding (`decodeLin`, transport lemmas) are fully removed.
- **Exact dedup key in domainBatch** — the one intentional behavior change: the candidate-set key is now a sorted duplicate-free `VarId` list instead of concatenated name strings, fixing collisions between equal-named variables with different powdr IDs (adversarial inputs only; compile-time regression guards included; no benchmark circuit is affected).
- **Legacy tree deleted** — the `Variable`-based pass implementations (`OldVariableBased/`, 42 files), the `ofSpec` decode/re-encode on-ramp, and the orphaned `Variable`-side support layer are removed; pass authoring is native-only (`DenseVerifiedPassW.of` / `.ofExtending`), with AGENTS.md and the architecture doc updated to match.
- **Hygiene** — comments/docstrings describe the code as it stands; migration-era narration is gone.

Verification: per-case circuit sizes are byte-identical to `main` across all benchmark corpora (CI effectiveness delta exactly +0.000 on every axis at every step of the series), proof integrity rests on exactly Lean's three standard axioms, and whole-corpus runtime improves consistently (openvm-eth quick bench: −9% to −20% vs `main` across runs, on top of the earlier dense-cleanup milestone's 16–30% corpus-wide gains).


Main.lean
```diff
@@ -28,6 +28,7 @@ provably terminating on the lexicographic size key `(vars, bus, constraints)`.
 -/
 
 open ApcOptimizer.OpenVM
+open ApcOptimizer.Dense
```
georgwiese, line 31: Why does Main have to change? The interface to the optimizer should be the same? I think it is the profiling stuff?

Maybe we should (not in this PR) have another binary for profiling and whitelist it for agents to change. Would be good if the main binary was simple and doesn't need changes.

georgwiese (commented):
Cool, looks like it improved a lot! Are there any more TODOs?

Main.lean
pacheco: Yeah, its basically the profiler.
One reason is the new "dense" pass interface.
Another is that it is measuring the cost of encoding/decoding to Variable=>VarId

PR #167, opened 2026-07-22
monicScale: evaluate l.norm once per affine factor

pacheco:
## Summary

Runtime-only fix for item 1 of the pass-improvement findings from the VarId migration audits: `denseMonicScaleAffine` recomputed the linearized form's normalization `l.norm` twice per constraint factor — once for the monic-gate test (`l.norm.terms`) and once for the rescale (`(l.norm.scale k⁻¹).toExpr`).

The normalization is now hoisted into a helper, `denseMonicScaleNormed`, which takes the normalized form as an argument, so it is evaluated once. A helper argument (rather than a `let` in the match) keeps the three `unfold` + `split` proof scripts in `MonicScaleProof.lean` working unchanged apart from unfolding the new name.

## Output-neutrality

The change is pure hoisting — no behavior change. Verified locally: `opt-export` output is byte-identical (`cmp`) on openvm-eth cases 001, 003, 010, 025, 050, 075, 100. `lake build` completes warning-free.

While auditing this file, two further output-neutral runtime inefficiencies were spotted in `denseMonicScale` (its recursive results are written out three times each — no CSE, so 3^depth recursive work — and accepted `.mul` factors are linearized twice); they are recorded in the local findings notes as follow-up candidates, not included here.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/MonicScale.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/MonicScale.lean
@@ -36,17 +36,22 @@ variable {p : ℕ}
 
 /-! ## Scaling with a unit certificate -/
 
+/-- The core of `denseMonicScaleAffine`, over the already-normalized linear form `n` (taking it
+    as an argument evaluates the normalization once for both its uses below). -/
+def denseMonicScaleNormed (e : DenseExpr p) (n : DenseLinExpr p) :
+    DenseExpr p × ZMod p × ZMod p :=
+  match n.terms with
+  | (_, k) :: _ =>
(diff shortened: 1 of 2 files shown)


pacheco:
Closing: profiling across all 303 local benchmark cases (openvm-eth, wasm-eth, keccak, sp1/rsp, sp1/keccak) plus a large sha256 APC shows monicScale is a cold pass — 95 ms cumulative in the worst case (wasm-eth apc_012), typically under 20 ms, ~0.1% of any case's runtime. The hoist is correct and output-neutral (CI confirmed identical circuits corpus-wide), but not worth the churn. The apparent sp1/rsp +58% in the first quick-bench run was runner noise: a rerun showed -0%, and a same-runner Runtime Bench plus local interleaved A/B both showed 1.00×.

PR #177, opened 2026-07-22
Simplify the implementation: generic pass builders, proof-transport factoring, dedup (output-identical)

georgwiese:
Maintainability pass, continuing #171/#174/#175: fewer pass wirings, generic builders instead of per-pass boilerplate, one shared proof transport where two parallel re-derivations lived, and one byte-check classifier where three recognizer twins lived. Code net **−401 lines** vs main (excluding the append-only `docs/log.md`), at **byte-identical runtime output**: the CI matrix confirms per-case circuit sizes identical to main on all five benchmark sets, with the report-only runtime row at or slightly below main everywhere (openvm-eth −5%).

## 1. Generic entailed-check skeleton (`EntailedCheck.lean` + `Proofs/EntailedCheck.lean`)

Several passes shared one shape: recognize a stateless check whose acceptance is exactly the vanishing of an entailed constraint, append those constraints, drop the recognized checks. That shape is now a reusable skeleton:

- `DenseCheckRule` — a recognizer bundled with its four obligations (unit multiplicity, statelessness, variable containment, acceptance ⟺ constraint vanishes) — and `DenseVerifiedPassW.ofCheckRules`, which turns a rule list into the append-and-drop pass (grouped per recognizer, provably the same output as running the rules as consecutive passes).
- `DenseVerifiedPassW.ofAddConstraints` — the append-only sibling: emitted-constraint list + containment/entailment obligations gives a pass.
- The reusable `DensePassCorrect.denseAddConstraints` / `denseFilterBusEntailed` move here from their historical homes in `Proofs/BusUnify.lean` / `Proofs/FlagFoldDrops.lean`.

Consequences: **`zeroWidthRange` + `rangeBool` merge into one `degenRange` pass** (their four files become `DegenRange.lean` + `Proofs/DegenRange.lean`, holding only recognizers and soundness lemmas); **`xorEqExtract`, `oneHotAnnihilate`, `zeroRegister`, `intervalForce`** rewire through `ofAddConstraints`, each dropping its bespoke covered/correct/wiring block; the two private `dpcRefl` copies of `DensePassCorrect.refl` are gone. A future pass of either shape is recognizers + soundness lemmas + one builder call — AGENTS.md and the architecture doc updated accordingly.

## 2. Same-env rewrite transport (`DensePassCorrect.ofEvalAgree`, Bridge.lean)

The two `domainFold` correctness proofs (direct `denseFoldOutV` and in-place `denseFoldOutInPlaceV`) were parallel ~145-line re-derivations of one argument: a pass that maps bus interactions by an eval-preserving rewrite and whose constraints vanish exactly when the input's do is correct with the identity environment. That transport is now proved once in Bridge; each fold proof is a ~45-line instantiation supplying only its membership arguments and the covered-constraints anchor. Proof-only — runtime definitions untouched.

## 3. One byte-check classifier (`denseByteShape?`, ByteCheckPack)

`denseSvCheck?` (ByteCheckPack), `denseByteCheckOperands?` (RedundantByteDrop) and `denseMatchByteSingle` (TupleRange) matched the same `byteXorSpec` shape family — XOR self-check, zero-operand mirrors, NOT/255 forms, OR-identity mirrors, packed pair — with near-identical ~110-line twin branch proofs. One classifier, parameterized on the constant-comparison predicate (structural vs `constValue?`-folded) and preserving each consumer's exact branch order, gates and multiplicity policy, now carries the single per-shape soundness proof; the recognizers are thin projections with their soundness lemmas' statements unchanged. The triplicated byte-complement arithmetic (`val_255_sub`/`val_255`) is also down to one copy.

## Tried and reverted: digitFold removal

A removal probe (13-case local sample) suggested `digitFold` was fully covered by the surrounding passes; the full CI matrix then showed **8 of 100 openvm-eth cases regress** (+3..+11 vars each) while the other four sets stayed identical — so the pass stays, and the probe→CI lesson is in `docs/log.md` entries 124–125. Also probed and kept (sample regressions): carryBranch, zeroRegister, intervalForce, xorEqExtract, zeroWidthRange. Consolidations analyzed but deliberately deferred (worse risk/benefit) are listed in `docs/ideas.md` → "Structural follow-ups".

## Verification

- `lake build` warning-free at every commit; proof-integrity + unused-theorem linter green (correctness axioms `{propext, Classical.choice, Quot.sound}`-only).
- Every kept change is byte-identical on a 13-case local sample (5 openvm-eth, 5 sp1/rsp, 3 wasm-eth); the CI matrix verdict on all five sets: **per-case circuit sizes identical**.

https://claude.ai/code/session_01UX4hWEyuenmgn68KFeCAZf


georgwiese:
Cool! LGTM, but letting @pacheco merge it, if it doesn't invalidate his concurrent work.

PR #181, opened 2026-07-22
Runtime: reverse Gauss sweeps to avoid fill-in

pacheco:
## Summary

- run both Gauss elimination sweeps in reverse algebraic-constraint order
- adapt the Gauss loop invariant proof using membership preservation under `List.reverse`
- record the local runtime experiment and leave corpus effectiveness to CI

## Why

Gauss elimination order controls transient fill-in. On `OpenVM/wasm-eth/apc_036_pc0x224240`, forward order grew the solution map to 111,005 terms before a late 500-pivot block rewrote 60,976 older rows over 14.35 million input terms. Reversing the sweep reaches those bridge variables first and avoids the rewrite cascade.

## Local impact

Two full reverse-order profiles:

- whole optimizer: 46.944 s and 46.727 s, versus 84.826 s on `origin/main`
- Gauss: 3.748 s and 3.760 s, versus 42.064 s
- about 91% less Gauss time (11.2x) and 45% less end-to-end time (1.81x)
- eight cleanup iterations in every run
- every reported cycle size matched; final size remained 1954 variables / 1201 bus interactions / 1895 constraints

The full benchmark CI should determine whether fixed reverse order preserves or improves effectiveness and runtime across all benchmark families.

## Validation

- `lake build`
- `./Scripts/check-proof-integrity.sh`
- two complete local `profile` runs on wasm-eth apc_036

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/Gauss.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/Gauss.lean
@@ -509,25 +509,26 @@ theorem denseTrySolveUnit_vars_subset (l : DenseLinExpr p) (v : VarId) (w : VarI
 
 /-- Batch linear (Gauss) elimination. From a constraint like `x - 2*y - 3 = 0` it derives the
     assignment `x := 2*y + 3` and substitutes it everywhere, dropping `x`. The cheapest solvable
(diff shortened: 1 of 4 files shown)


pacheco:
not merging this one yet, seems just to be a heuristic that works for some specific APCs

pacheco:
#190 solves this better

PR #195, opened 2026-07-23
docs: render the audited surface as a Verso site

georgwiese:
Renders the audited surface of `apc-optimizer` as a single-page, cross-linked [Verso](https://github.com/leanprover/verso) site under `docs/`. Prose is interleaved with the real definitions and theorem statements, spliced live from the compiled source via `{docstring}` and source `anchor` regions — so the formal content shown *is* the machine-checked artifact, not a transcription. Hover any name for its type; follow any link to its definition.

### What it covers
A guided reading of the whole audited surface: expressions and circuits → bus interactions and semantics (incl. the memory discipline, Blum et al.) → the asymmetric soundness/completeness relation (side effects, derivations/witgen, admissibility) → the degree bound → the top-level `Optimizer.isCorrect`. References are cited inline as numbered margin notes.

### How it's wired
- A **non-default** lake target (`Docs` lib + `docs` exe under `docs/`), so `lake build` and the audited CI never build it; it shares the main workspace, reusing mathlib's cache.
- `docs/build.sh` emits a self-contained single-page site; `--serve`/`--watch` build and serve locally with live reload (see `docs/README.md`). Serving over HTTP is required — hovers/search fetch JSON at runtime.
- `.github/workflows/deploy-docs.yml` builds the site on every relevant PR (a spec/docs drift check — a rename or moved `-- ANCHOR:` fails the build) and deploys to GitHub Pages on `main`.

### Spec.lean
Only documentation-facing edits: docstrings on `Variable`/`Expression`/`ComputationMethod` fields, `-- ANCHOR:` regions for the spliced blocks, and a bound-variable rename (`env` → `assignment`). No definition or theorem statement changes.

### Verification
- `lake build` (audited targets) and `lake build docs` both green.
- `Scripts/check-proof-integrity.sh` passes: correctness theorems rest only on the three standard axioms; no forbidden tactics; no unused theorems.


lakefile.toml
```diff
@@ -2,14 +2,35 @@ name = "apc-optimizer"
...
+git = "https://github.com/leanprover/verso"
+rev = "v4.30.0-rc1"
+
+# Kept last so its versions of transitively-shared dependencies win. Pinned to an
+# exact rev (not "master") so `lake update` cannot drift it.
 [[require]]
 name = "mathlib"
 git = "https://github.com/leanprover-community/mathlib4"
-rev = "master"
+rev = "e60d89a05065d2697d54edf3ca85f05ca920bbb4"
```
leonardoalt: will this be limiting for code in the future?

leonardoalt:
@codex review

leonardoalt:
@codex review

docs/Docs.lean
```diff
@@ -0,0 +1,280 @@
...
+several thousand lines of optimization passes that implement it need no audit, because each pass
+carries its own correctness proof and the master theorem composes them.
+
+# Circuits
+
+A circuit is a list of algebraic constraints together with a list of bus interactions, both over
+arithmetic expressions in a finite field. The algebraic constraints are the AIR-style identities of
+STARK arithmetization {citep stark}[]; the bus interactions are the multiplicity-weighted messages
+of a logarithmic-derivative lookup argument {citep logup}[], the mechanism modern zkVMs use for
+range checks, table lookups, memory, and inter-chip communication.
```
leonardoalt: "chip" has not been mentioned/defined before.

docs/Docs.lean
```diff
@@ -0,0 +1,280 @@
...
+
+An {deftech}_expression_ is a constant, a variable, or a sum or product of expressions:
+
+{docstring Expression}
+
+Expressions are evaluated under an {deftech}_environment_ assigning a field element to every
+variable, and carry a multiplicative {deftech}_degree_ — the quantity the proving backend bounds.
```
leonardoalt: maybe clarify that this is an upper bound for circuit/gen apcgen?

docs/Docs.lean
```diff
@@ -0,0 +1,280 @@
...
+
+{docstring Expression.eval}
+
+{docstring Expression.degree}
+
+A {deftech}_bus interaction_ sends a payload tuple to a global bus with a multiplicity. Buses are
+how a chip talks to the rest of the system: a stateless bus is a lookup (a chip sends with
```
leonardoalt: again, "chip" undefined

docs/Docs.lean
```diff
@@ -0,0 +1,280 @@
...
+multiplicity 0/1, a table chip receives), while a stateful bus (memory, the execution bridge)
```
leonardoalt: it might be more confusing than not to say that a chip sends with mult 0, reader might immediately ask "well why would it send 0 in the first place", which then takes longer to explain.

docs/Docs.lean
```diff
@@ -0,0 +1,280 @@
...
+
+## What a circuit does
+
+A circuit is {deftech}_satisfied_ under an environment when every algebraic constraint vanishes and
+no active bus interaction violates another chip's table:
+
+{docstring ConstraintSystem.satisfies}
+
+An assignment is {deftech}_admissible_ when its active stateful messages satisfy the VM's
+`admissible` predicate — the real-trace assumption, lifted to a constraint system:
```
leonardoalt: is it a real-trace "assumption" or "requirement"?

docs/Docs.lean
```diff
@@ -0,0 +1,280 @@
...
+its output is both a sound and a complete replacement, and it respects the bound:
+
+{docstring Optimizer.isCorrect}
+
+# The memory discipline
+
+Memory and the execution bridge are stateful buses whose `admissible` predicate is the offline
+memory-checking discipline of {citet blum}[], reused in essentially every zkVM and, in its modern
+multiplicity-based form, by arguments like Twist-and-Shout {citep twistShout}[]. `apc-optimizer`
+does not prove this discipline; it *assumes* it about real traces (a completeness-only assumption)
```
leonardoalt: is it completeness-only? Don't we use this assumption in optimization passes?

docs/Docs.lean
```diff
@@ -0,0 +1,280 @@
...
+{docstring ApcOptimizer.SP1.Sp1BusType}
+
+{docstring ApcOptimizer.SP1.violates}
+
+{docstring ApcOptimizer.SP1.sp1BusSemantics}
+
+# The theorems
+
+The master theorem states that the optimizer is correct for *every* bus semantics and *every* choice
+of proven bus facts (the mechanism by which passes learn sound properties of the tables, itself
```
leonardoalt:
```suggestion
of proven bus facts (the mechanism by which optimization passes learn sound properties of the tables, itself
```

docs/Docs.lean
```diff
@@ -0,0 +1,280 @@
...
+And the SP1 optimizer:
+
+{docstring ApcOptimizer.SP1.sp1Optimizer_maintainsCorrectness}
+
+# Trust boundary
+
+![Trust map: green nodes are machine-checked and discharged by the proofs; amber nodes are what the auditor must establish by hand — the VM semantics, the memory discipline, and the input-circuit assumptions.](trust.svg)
+
+Everything reachable from the four theorems above is machine-checked and, per the project's
+CI, rests only on Lean's three standard axioms — no `sorry`, `axiom`, or `native_decide`. What an
```
leonardoalt: not sure we need to mention CI here, implementation detail

docs/Docs.lean
```diff
@@ -0,0 +1,280 @@
...
+
+> *The input guarantees invariants and respects the degree bound.* The optimizer *preserves*
+> `guaranteesInvariants` and `withinDegree`, but only assuming the input has them (e.g. that written
+> memory limbs are byte-range-checked).
+
+> *PC lookups are pinned.* The semantics checks only the *arity* of a PC lookup, not the program
+> table. We assume constraints fixing the lookup fields to constants have already been added to the
+> input, so the optimizer cannot alter them.
+
+> *Every memory writer is byte-range-checked.* The semantics treats a memory *receive* from the
```
leonardoalt: Is this true for SP1?

leonardoalt (commented):
Nice!
One general thing I dislike in the style is the countless use of the pattern " text ... - some remark - ... text". I think this is a bad prose pattern.

georgwiese:
Thanks for the review! Addressed in 50d6aaa:

**Prose style** — removed the "text — aside — text" em-dash pattern throughout (the doc now has zero em-dash asides), rewriting those as separate clauses/sentences.

**Inline comments (@leonardoalt):**
- *"chip" undefined* — "chip" is now introduced on first use, in the opening sentence of §Circuits (a circuit is the constraint system of a single chip).
- *degree upper bound* — clarified: expressions "carry a multiplicative degree, on which the proving backend imposes an upper bound," linking to §The degree bound.
- *"sends with multiplicity 0/1"* — dropped; a stateless bus is now described as "a chip looks a tuple up in a table … every lookup must be matched by a table entry," no confusing 0-multiplicity.
- *assumption vs requirement (admissible)* — reframed: "Completeness is required only for admissible assignments, so this is the condition that picks out the real traces the optimizer must preserve."
- *memory discipline "completeness-only" / used in passes* — reworded to "takes it as an assumption about real (admissible) traces, and the `busUnify` pass relies on it to cancel and chain memory accesses." Dropped "completeness-only."
- *"passes learn" suggestion* — applied ("optimization passes learn").
- *CI mention in trust boundary* — removed; now names the three axioms (`propext`, `Classical.choice`, `Quot.sound`) directly.
- *byte-range true for SP1?* — the memory-range assumption is generalized: "byte range for OpenVM, 16-bit limbs for SP1."
- *lakefile pin — limiting for future code?* — good catch. Reverted mathlib to `rev = "master"` (it tracked master before). The plausible/`cache get` fix comes from ordering mathlib **last** so its shared-dep versions win, not from pinning; the manifest's mathlib entry now matches main exactly (`inputRev: master`, same locked rev).

**@chatgpt-codex-connector (SP1 unverified lookup tables):** the "Lookup-table messages are pinned" assumption now explicitly lists SP1's instruction-fetch and page-protection lookups alongside the PC lookup as accepted on arity alone.

georgwiese:
@codex review

georgwiese:
Merging now, because I want to test deployment. Will likely do a follow-up PR to clean up a few things.

PR #206, opened 2026-07-24
perf(domainFold): Array.modify accepted folds at touched positions — pass 173s → 6.9s on 170k sha256

georgwiese:
## What

Stacked on #205. `denseFoldLoopV` threads the list system `d` and re-materializes it per accepted fold: `denseFoldOutIdxV` maps over `algebraicConstraints.zipIdx` / `busInteractions.zipIdx` (each an O(system) `toArray` + full rebuild) and `DenseFoldIdx.refresh` converts the folded lists back to arrays. On sha256-scale systems the `List.toArray` length pre-pass alone was ~8% of total optimizer CPU (`l_List_lengthTR`), plus the map rebuild and the RC churn of freeing the old spine.

The new `denseFoldLoopArrV` threads only the (already array-backed) `DenseFoldIdx`, applies an accepted fold with `Array.modify` at the touched (bucketed) positions — O(touched) per accept — and materializes lists once at the pass exit.

## Proof structure

The list loop stays as the specification. `denseDomainFoldFV_eq_fast` proves the array twin equal (via a generic `foldl`-`Array.modify` = sparse positional map lemma) and installs it with `@[csimp]`, so the existing pass-correctness proofs are untouched.

## Measured (sha256_big, 170k vars, `profile`, stacked on #205)

- domainFold: **172 784 ms → 6 933 ms** (~25×)
- reencode also dropped 264s → 178s from the reduced allocator/RC pressure (unchanged expressions now stay physically shared)
- **total: 989s → 643s (−35%)** for the #205+#206 pair

`Scripts/check-proof-integrity.sh` passes (the equality is one of the checked `@[csimp]` roots); `run` output byte-identical across all openvm-eth fixtures and on ecrecover_53k (indexed path).

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/DomainFoldRuntime.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/DomainFoldRuntime.lean
@@ -249,6 +249,51 @@ def denseFoldLoopV : List (List VarId) → DenseConstraintSystem p → DenseFold
     let r := denseFoldStepV d fidx xs
     denseFoldLoopV rest r.1 r.2
 
+/-! ## The array-native indexed fold loop (runtime twin)
+
+`denseFoldLoopV` threads the *list* system `d` and re-materializes it per accepted fold:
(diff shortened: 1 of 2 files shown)


georgwiese:

PR #207, opened 2026-07-24
perf(busUnify,busPairCancel): linear busId dedup

georgwiese:
## What

Stacked on #206. Both `denseBusUnifyF` and `denseBusPairCancelPass` deduplicate the per-interaction busId list with `List.dedup` — an **O(n²)** membership walk over the system-sized interaction list (71k interactions in sha256_big's first cleanup cycle), re-run every cleanup cycle by both passes.

`HashedDedup.hashedDedup` is proven identical (`hashedDedup_eq : hashedDedup hf l = l.dedup`) and buckets membership tests by hash, so the walk is linear. The structural twin `denseBusUnifyNew` in `Proofs/BusUnify.lean` is updated in lockstep (it must mirror the transform's body).

## Expected effect

Sum over cycles of (#interactions)² comparisons disappears from **both** passes (busPairCancel 106s, busUnify 51s on 170k sha256 before this change). Exact numbers will show in the sha bench; local measurement to follow in a comment.

Output is provably unchanged (`hashedDedup_eq`); `run` byte-identical across all openvm-eth fixtures; `Scripts/check-proof-integrity.sh` passes.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/BusPairCancel.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/BusPairCancel.lean
@@ -355,7 +355,9 @@ def denseCancelLoop (cs0 : DenseConstraintSystem p) (bs : BusSemantics p) (facts
 def denseBusPairCancelPass (pw : PrimeWitness p) (aggressive : Bool) : DenseVerifiedPassW p :=
   fun reg d hcov bs facts =>
   if hp1 : (1 : ZMod p) ≠ 0 then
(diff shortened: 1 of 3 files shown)


georgwiese:
Measured on sha256_big stacked on #205+#206: **no change** (busPairCancel 106.4→108.2s, busUnify 51.4→52.5s — noise). `List.dedup` compares each element against the *kept* accumulator, which here holds only the ~6 distinct bus ids, so it was already O(n·distinct), not O(n²) — my assumption was wrong. Closing: the asymptotic tidy-up isn't worth the review, and I'll restack #208 directly onto #206.

PR #208, opened 2026-07-24
perf(pointwiseDupDrop): slot-0-indexed sweep representatives + hoisted singles in the verifier

georgwiese:
## What

Stacked on #207. Two independent rescans made flagFold's pointwise-duplicate drop expensive on range-check-heavy systems:

1. **Sweep**: an incoming interaction was tested against *every* representative in its `(busId, mult, payloadLen)` class (`densePdSigsCompatible` was 3.4% of total CPU). A certified twin's slot pair must be value-equal or share a decomposition carrier (`denseSlotEqCert`), so every true match has a representative whose **first payload slot** is hash-equal to — or shares a variable with — the incoming slot 0. Representatives are now indexed by slot-0 value hash and per-variable buckets, and only those candidates are re-checked with the same full tests. The proposed drop set is unchanged, and it remains a heuristic *proposal* that `densePdKeep` re-verifies, so no proof is touched.

2. **Verify**: the re-verification spelled `denseSingleVarCs d.algebraicConstraints` out inside the `filterMap` closure, recomputing the O(system) single-variable scan (hashing every constraint's variable list) **once per proposed drop**. It now reuses the already-let-bound `singles`.

## Verification

`run` byte-identical across all openvm-eth fixtures vs the base branch; `Scripts/check-proof-integrity.sh` passes. flagFold was 85s of the 643s sha256_big profile at the base of this stack; measurement comment to follow.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/FlagFoldDrops.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/FlagFoldDrops.lean
@@ -1,5 +1,6 @@
 import ApcOptimizer.Implementation.OptimizerPasses.FlagUnify
 import ApcOptimizer.Implementation.OptimizerPasses.HashedDedup
+import ApcOptimizer.Implementation.OptimizerPasses.SearchBudgets
 
 set_option autoImplicit false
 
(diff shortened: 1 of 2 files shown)


georgwiese:
Measured on sha256_big (170k vars, `profile`, quiet 48-core box), stacked on #205+#206 (643s total baseline):

- flagFold: **85.0s → 15.6s**
- total: **643s → 581s**

Combined stack so far vs yesterday's main: 1298s → 581s (**2.2× faster**). Output byte-identical on all openvm-eth fixtures.

georgwiese:
Re the +51% on openvm-eth: the same CI comment's per-pass table shows openvm-eth pass-for-pass parity (worst 1.04×), and locally the full stack measured +2.5% wall on 50 fixtures — so most of that number looks like runner variance. The +2.5% was real overhead though: the slot-0 index costs two pair-keyed map lookups per interaction, which doesn't pay off on fixture-sized systems. Pushed a gate: systems under `pointwiseDupDropIndexThreshold = 4096` interactions keep the direct per-class scan (both paths propose the identical drop set, so effectiveness is unchanged). Local re-check: 50 fixtures 98.1s (base) / 100.6s (ungated) / 97.1s (gated); sha256_big flagFold unchanged at 15.6s (all cycles that matter are above the gate). Also rebased onto main after #205.

georgwiese:
Post-gate CI: the openvm-eth wall row still shows +15%, but the per-pass table in the same comment is at exact parity (reencode 1.00×, flagFold 0.99×, all others ≤1.04×) and effectiveness is identical — the wall row on the parallel matrix runners appears to carry ~±15-50% noise between runs. sha256 runtime row: **832.6s → 598.5s (−28%)** with #205 now in main on both sides.

georgwiese:

PR #209, opened 2026-07-24
perf(rootPairUnify): serve bound queries from per-variable indexes

georgwiese:
## What

Stacked on #208. `denseAnyVarBound` is queried per candidate pair with the *full* lists: `denseFindVarBound` walks every interaction, and on failure `denseScaledSlotBound` walks every interaction again, resolving each offset variable's finite domain with an O(constraints) `denseFindDomainAlg` scan (`DenseExpr.mentions` subtree tests over the whole system were ~3% of total CPU). rootPairUnify was 49.6s of the 581s sha256_big profile at this stack's base.

The pass now builds, once per invocation:
- a per-variable interaction bucket (`denseVarBucket`, moved to a shared `VarBucket` module so both `RedundantByteDrop` and `RootPairUnify` can use it), and
- a tabulated first-yield domain map (`denseFindDomainMap`),

and serves the same queries from those.

## Proof structure

An interaction can only bound `x` if it mentions `x` (a bounding slot is literally `.var x`; an accepted scaled slot has a nonzero `splitAt` coefficient), and the per-variable bucket is exactly the mention-filtered sublist in source order — so every query answer is unchanged. `denseRootPairUnifyF_eq_fast` proves the pass equal and installs the indexed body via `@[csimp]`; the soundness proofs stay on the scanning specification.

## Verification

`run` byte-identical across all openvm-eth fixtures and ecrecover_53k vs the base branch; `Scripts/check-proof-integrity.sh` passes. sha256_big measurement to follow in a comment.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/Proofs/RootPairUnify.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/Proofs/RootPairUnify.lean
@@ -518,6 +518,514 @@ theorem denseRootPairUnifyF_correct (pw : PrimeWitness p) (reg : VarRegistry) (b
       (fun i t hti z hz => hinv.2 i t hti z hz)
   · exact DensePassCorrect.refl reg.isInput d bs
 
+/-! ## The indexed twins compute the same pass (`@[csimp]` runtime replacement)
+
+`denseRpLoopIdx` (`RootPairUnify.lean`) serves each bound query from a per-variable interaction
+index and a tabulated first-yield domain map instead of scanning the full interaction and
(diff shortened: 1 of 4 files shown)


georgwiese:
Measured on sha256_big (`profile`, quiet box), stacked on #205+#206+#208 (581s baseline): rootPairUnify **49.6s → 33.7s**, total **581s → 564s**. The remaining 33.7s is in the candidate/two-root machinery itself (linearize + certificate checks per pair attempt), not the scans — that's ZMod-instance-reconstruction territory, a separate follow-up. Cumulative stack vs yesterday's 1298s: **2.3× faster**.

georgwiese:

PR #210, opened 2026-07-24
perf: 1.8x faster on large APCs — domainFold/flagFold/rootPairUnify/busPairCancel (989s → 550s on 170k sha256)

georgwiese:
Combined stack (one commit per change, formerly #206/#208/#209 + this PR's segment scans), targeting main directly. Each change keeps the optimizer's output **bit-identical** — either the fast body is proven equal to the specification and installed with `@[csimp]` (the pass proofs never move off the spec), or it is an untrusted proposal that the existing verified re-check consumes — and `Scripts/check-proof-integrity.sh` passes throughout.

## The changes

1. **domainFold: apply accepted folds with `Array.modify` at touched positions** (`denseFoldLoopArrV`, csimp twin). Previously every accepted fold re-materialized the whole system (`zipIdx`'s `toArray` + full-list map + `refresh`'s `toArray`) — `List.toArray`'s length pre-pass alone was ~8% of total runtime. **173s → 6.9s** on the pass, and reencode drops 264s → 178s untouched, purely from allocator/RC pressure.

2. **pointwiseDupDrop: slot-0-indexed sweep representatives + hoisted `singles`.** A certified twin's slot pair is value-equal or shares a decomposition carrier, so only representatives hash-equal to — or variable-sharing with — the incoming slot 0 are re-checked (identical drop set, still re-verified by `densePdKeep`); the verifier no longer recomputes the O(system) single-variable filter per proposed drop. Gated to systems ≥ `pointwiseDupDropIndexThreshold = 4096` interactions (fixture-scale systems keep the direct scan). **flagFold 85s → 15.6s.**

3. **rootPairUnify: bound queries served from per-variable indexes** (`denseVarBucket` in the new shared `VarBucket.lean` + tabulated first-yield domain map, csimp twin; an interaction can only bound `x` if it mentions `x`, so answers are unchanged). The builds are thunked, so systems without two-root twins never pay for them. **49.6s → 33.7s.**

4. **busPairCancel: candidate region checks fold the array segments directly** (`denseLiveAllSeg`/`denseShieldScanSeg` with proven `…_eq` lemmas) instead of materializing an O(prefix) list per candidate pair. **107.4s → 89.5s.**

Plus `agent-docs` log entries 137–141 and a runtime-ideas refresh.

## Measured (sha256_big, 170k vars, powdr dev server, quiet)

**989s (main) → 550s (−44%)**; with #205 (already merged) the two-day total is 1298s → 550s (**2.4×**). Local A/B on the other sets: openvm keccak 36s → 30s (−16%), sp1 keccak 97.3s → 97.3s (±0), openvm-eth fixtures at parity post-gate (98.1s vs 97.1s over 50 cases). Tip-of-stack CI: parity-or-better on all six sets (openvm-eth −0%, wasm-eth −12%, keccak −17%, sha256 −31%, sp1/rsp −33%, sp1/keccak −1%); the wild runtime rows on the intermediate PRs' runs (+55% sp1/rsp on a subset branch vs −33% on its superset) are runner noise — per-pass tables show ≤1.04× throughout. A serial local suite comparison is posted below.

## Verification

- `run` output byte-identical vs base across all 100 openvm-eth fixtures and ecrecover_53k at every step.
- `Scripts/check-proof-integrity.sh` green (three standard axioms only; csimp equalities among the checked roots).

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/BusPairCancel.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/BusPairCancel.lean
@@ -202,18 +202,16 @@ def denseFindCancelGoIdx (cs0 : DenseConstraintSystem p) (bs : BusSemantics p) (
             by_contra hc
             rw [Array.getElem?_eq_none (Nat.le_of_not_lt hc)] at hR; simp at hR
           have hSget : arr[i]? = some S := Array.getElem?_eq_getElem hi
-          let B := denseLiveArr arr alive hsz (i + 1) (j - i - 1) (by omega)
-          if hmidB : B.all (denseMidRefuted ops shape T busId S) = true then
-          let A := denseLiveArr arr alive hsz 0 i (by omega)
-          if hshieldA : denseShieldOk ops shape T busId S A = true then
-          have hBeq : B = denseLiveSeg arr alive (i + 1) (j - i - 1) :=
-            denseLiveArr_eq arr alive hsz (i + 1) (j - i - 1) (by omega)
-          have hAeq : A = denseLiveSeg arr alive 0 i := denseLiveArr_eq arr alive hsz 0 i (by omega)
+          if hmidB : denseLiveAllSeg arr alive (denseMidRefuted ops shape T busId S)
+              (i + 1) (j - i - 1) = true then
+          if hshieldA : (denseShieldScanSeg ops shape T busId S arr alive 0 i).2 = true then
           have hmid : ∀ m0 ∈ denseLiveSeg arr alive (i + 1) (j - i - 1),
               denseMidRefuted ops shape T busId S m0 = true := by
-            rw [← hBeq]; exact fun m0 hm0 => List.all_eq_true.mp hmidB m0 hm0
+            rw [denseLiveAllSeg_eq] at hmidB
+            exact fun m0 hm0 => List.all_eq_true.mp hmidB m0 hm0
           have hshield : denseShieldOk ops shape T busId S (denseLiveSeg arr alive 0 i) = true := by
-            rw [← hAeq]; exact hshieldA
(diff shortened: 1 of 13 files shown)


georgwiese:
Measured on sha256_big (`profile`, quiet box), stacked on #205+#206+#208+#209 (564s baseline): busPairCancel **107.4s → 89.5s**, total **564s → 550.5s**. The remaining 89s is the per-candidate certificate work itself (the shield still evaluates `densePreRefuted` over the whole live prefix per candidate, O(live²) certs per invocation) — an incremental shield structure would be the next, much deeper cut. Cumulative stack vs today's main: 989s → 550s (−44%); vs yesterday's 1298s: **2.4×**.

georgwiese:
### Local runtime comparison — measured on the powdr dev server

Serial, quiet 48-core box, one `run` sweep per set per binary; main = `7a76d61` (includes #205), PR = `aa6139e`. Per-case circuit sizes are **identical** between main and this branch on every set (proven equalities / verified re-checks, plus byte-compared `run` output on all openvm-eth fixtures and ecrecover_53k).

| set | main | this branch | Δ |
|---|---|---|---|
| openvm-eth (100 cases) | 105.3 s | **98.4 s** | −7% |
| wasm-eth (100 cases) | 209.4 s | **207.3 s** | −1% |
| keccak (1 case) | 35.2 s | **28.2 s** | −20% |
| sha256 (1 case) | 858.9 s | **538.4 s** | −37% |
| sp1/rsp (100 cases) | 69.0 s | **68.8 s** | −0% |
| sp1/keccak (1 case) | 96.5 s | **95.2 s** | −1% |

No set regresses. The earlier ±30–55% swings on the per-PR CI runs were runner noise (superset branches measured faster than their subsets on the same sets; per-pass tables showed ≤1.04× throughout).

PR #213, opened 2026-07-26
Document optimizer scaling analysis and add R13–R15 to ideas.md

georgwiese:
## Summary

Add comprehensive documentation of the Lean optimizer's superlinear scaling behavior relative to circuit size, with cross-implementation comparison to powdr's Rust optimizer and detailed per-pass attribution. Integrate findings into the runtime section of `ideas.md` with three new ranked fix entries (R13–R15).

## Changes

- **New file: `agent-docs/runtime-scaling.md`** — A detailed session note (2026-07-26) analyzing why the Lean optimizer scales as ~N^1.5–1.8 in circuit size while powdr's Rust optimizer is ~N^1.0:
  - §1: Cross-implementation gap from reference benchmark data (20,115 APCs)
  - §2: Controlled scaling experiment using k disjoint copies of a single APC, isolating per-pass cost from fixpoint rounds (exponent 1.77)
  - §3: Per-pass attribution on sha256 (168,915 vars) and ablation showing cost is shared between passes, not owned by one
  - §4: Three recurring patterns producing superlinearity, with three previously-unlisted quadratic sites
  - §5–8: Why Rust avoids this, reproduction instructions, post-#211 re-measurement, and a four-tier roadmap to remove superlinearity

- **Modified: `agent-docs/ideas.md`** — Updated runtime section:
  - Added cross-reference to `runtime-scaling.md` with brief summary
  - Added **R13** (finish candidate-group family: `domainFold` and `denseCheckReencode` gates)
  - Added **R14** (`subsumedRange`/`subsumedCheck` bound index reuse, exponent 2.24)
  - Added **R15** (`bytePack`/`tupleRange` resume cursor and `flagUnify` domain tabulation, exponent 2.10)

- **Modified: `agent-docs/README.md`** — Added note that runtime analysis is documented in `runtime-scaling.md`

## Key Findings

The superlinearity is not a port defect but a structural difference in proof obligations: the Lean optimizer re-verifies each candidate against the full system (Θ(candidates × system) work), while Rust's single sweep applies all drops in one batch. The cost is concentrated in six passes (`reencode`, `busPairCancel`, `busUnify`, `domainFold`, `bytePack`, `flagUnify`), with `reencode` alone accounting for 38 % of sha256's runtime at ff15446 (reduced to 23 % by #211's indexing).

The disjoint-replica ladder is the key measurement instrument: it holds the fixpoint round count constant while scaling circuit size, making per-pass exponents visible and distinguishing genuine quadratics from constant-factor problems.

https://claude.ai/code/session_01NdrwAL7S5jwmeibazAxwcd

### DIFF
--- a/agent-docs/README.md
+++ b/agent-docs/README.md
@@ -10,6 +10,9 @@ documents — not part of the audited surface, and not a substitute for it.
   don't fit the current session and remove ones that have been implemented.
 - [`log.md`](log.md) — an append-only history of optimizer changes: each entry records the idea,
   whether it worked, and its effect on effectiveness. Earlier entries describe superseded designs.
+- [`runtime-scaling.md`](runtime-scaling.md) — why the optimizer's wall time is superlinear in
+  circuit size (~N^1.5–1.8 against Rust's ~N^1.0), with the measurements, the per-pass attribution,
+  and the quadratic code sites. Read before working on runtime.
 
 For humans, the entry points are the top-level [`README.md`](../README.md) (which defines the
 auditing surface) and [`AGENTS.md`](../AGENTS.md) (the agent instructions).
--- a/agent-docs/ideas.md
+++ b/agent-docs/ideas.md
@@ -120,6 +120,11 @@ touching these files anyway):
 
 ## Runtime
 
+For the cross-implementation picture — Lean at ~N^1.5–1.8 against powdr Rust's ~N^1.0, the
+disjoint-replica scaling experiment that isolates per-pass cost from fixpoint rounds, and why the
+gap is a proof-obligation shape rather than a port defect — see
+[`runtime-scaling.md`](runtime-scaling.md).
+
 Rewritten 2026-07-18 from a fresh profiling session (per-pass `profile`, per-cycle timing, gdb
 stack sampling, and a size-scaling sweep — all on this container, serial). **Runtime is
 end-to-end quadratic in circuit size** (openvm-eth sweep: input 2.3k → 8.8k items costs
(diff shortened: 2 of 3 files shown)


georgwiese:
Closing — agreed, though for a different reason than the benchmark: this PR is docs-only (440 added lines, no `ApcOptimizer/` change), so the effectiveness/runtime tables compare a binary against itself and cannot show a benefit either way.

The real reason to close is maintenance. Its §7–§8 (post-#211 re-measurement and the four-tier roadmap) are already superseded by #216 and #217, and the rest is now duplicated by log entry 145, `ideas.md` R3 and `Benchmarks/scaling_bench.py`. A standalone 440-line document that has to be hand-refreshed after every runtime PR is a liability.

Two findings were not captured anywhere else, so they moved to the runtime section of `ideas.md` in `claude/reencode-sparse-accept` (commit `8f723ea`) before closing:

- **The target.** Median µs per input variable across the 20 115-APC comparison: Rust 1248 → 679 as circuits grow from <500 to 64k–300k variables, this optimizer 923 → 4083. Rust flat, we grow 4.4×. We are 1.3–2.6× *faster* below ~8k variables and 32 % faster summed over all APCs — the problem is confined to the large tail, which is also where a single call costs minutes.
- **Why the gap exists.** `redundant_memory_interactions_indices` is one left-to-right sweep with a live-address map and batched drops, never re-verifying the region between a send and its receive. We have that sweep (`denseSweepGo`), then re-verify every candidate it proposes, because that is what a machine-checked certificate requires. So: Θ(system) per candidate × Θ(N) candidates — a proof-shape difference, not a port defect.

---

PR #215, opened 2026-07-26
perf(addrDiseq): scope the two-root build to address slots, linearize factors once (404s -> 378s on 170k sha256)

georgwiese:
`DenseTwoRootMap.build` — the address-disequality certificate table `busUnify` and `busPairCancel`
consult — was ~4.9 % of whole-run CPU on the 168k-variable sha256 APC, rebuilt per pass invocation
over the **whole** constraint list. Two independent fixes, both output-preserving:

## What changed

- **Scope the build to what can be queried.** Only variables at an address-slot position of some
  declared memory shape are ever looked up: `denseAddrTwoRootNeq` reads
  `S.payload[slot]`/`m.payload[slot]` for `slot ∈ shape.addressFields`, and `densePtrReductions`
  keys on the queried form's own variables. `DenseTwoRootMap.buildForAddrs` indexes only the
  constraints mentioning such a variable (`denseAddrSlotVars` collects them across every
  interaction — the *candidate's* shape decides the slots, so an interaction on any bus can be read
  at another bus's address positions). Every reachable entry is unchanged: an entry for `v` is only
  inserted by a constraint mentioning `v` (`addVars` folds over `c.vars`), and such a constraint
  mentions an address variable, so it survives the filter. Soundness needed one lemma —
  `DenseTwoRootMap.Sound.mono`, since `Sound` only asserts *some* witnessing constraint in the list.
- **Linearize a product's factors once per constraint.** `denseTwoRootOf? c v` re-ran
  `denseLinearize` on both factors for every variable of `c`. `addVarsFast` hoists the two
  linearizations out of the per-variable loop and is wired in as the compiled twin
  (`@[csimp] addVars_eq_fast`), so every proof still sees `addVars`. This was the larger half.

## Impact

sha256_big (146k constraints / 71k interactions / 168k vars, `profile`, quiet 48-core box):

| | main (47c71eb) | this PR |
|---|---|---|
| busUnify | 54.1 s | **40.2 s** (−26 %) |
| busPairCancel | 50.1 s | **39.6 s** (−21 %) |
| total | 404.4 s | **377.6 s** (−6.6 %) |

Cumulative with #211 and #212 on the same box: **523.9 → 377.6 s (−28 %)**. wasm-eth apc_036
`opt-export`: 39 → 33 s.

Note on sizing: the address scoping *alone* was only 404.4 → 398.2 s, well under its 4.9 % CPU
share — sha256's bridge-bus address slots mention variables that appear in a large share of the
constraints, so the filter keeps more than expected. The hoisting supplied the rest.

## Validation

- `lake build` (warning-free), `Scripts/check-proof-integrity.sh` (three standard axioms, no unused
  theorems)
- byte-identical `opt-export` vs. main on openvm-eth apc_044 / apc_067, keccak apc_001, wasm-eth
  apc_036 and OpenVM sha256 apc_001 (the 168k-var case)
- `agent-docs/log.md` entry 144; `agent-docs/ideas.md` runtime section refreshed — reencode (96 s)
  is the top pass again, and its remaining cost is diffuse (survivor enumeration, per-target
  `c.vars.eraseDups` gates, per-accept index rebuilds)

https://claude.ai/code/session_01HKR42FVxb2GLyEesKGgyvK

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/AddrDiseq.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/AddrDiseq.lean
@@ -14,6 +14,15 @@ variable {p : ℕ}
 
 /-! ## Recognizing a two-root constraint (dense) -/
 
+/-- The two-root entry for `x` from a product's already-linearized factors. -/
+def denseTwoRootOfLins (l1 l2 : DenseLinExpr p) (x : VarId) :
+    Option (ZMod p × DenseLinExpr p × ZMod p) :=
+  let k := l1.coeff x
+  let A := (l1.others x).norm
+  let A2 := (l2.others x).norm
+  if k ≠ 0 ∧ l2.coeff x = k ∧ A2.terms = A.terms then some (k, A, A2.const - A.const)
+  else none
+
 /-- The two-root decomposition of a dense constraint relative to `x`: `some (k, A, δ)` when the
     constraint is a product of two affine factors, both linear in `x` with the same nonzero
     coefficient `k`, whose `x`-free parts differ by the constant `δ`. -/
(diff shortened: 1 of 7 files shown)


georgwiese:
Local validation on the big case (the CI sha job runs the small parallel set): byte-identical `opt-export` on the 168k-var OpenVM sha256 APC, that run 392 s → 365 s. CI confirms per-case circuit sizes identical on all six sets.

PR #234, opened 2026-07-28
perf: mine byte-operand bounds into the domain table (domainBatch ~4× on SP1)

qwang98:
Runtime-only, output-identical, fully proven. (Re-targeted to main: the prior #232 merged into the #219 branch, but #219 squash-merged to main without this change, so domainBatch never reached main. This is the same change, rebased onto current main.)

## Problem
domainBatch is the biggest SP1 optimizer cost. Bisection showed the table *build* is ~2%; **94% is box scans brute-forcing the full 2^16 domain of 16-bit memory limbs** (~38M points on SP1 keccak). Each limb is bounded <256 by a byte op via an affine operand `a·x+b`, but its domain was only the bare `.range 65536`.

## Fix
`denseAddByteDoms` mines each byte interaction's operand bound into the domain table as a ≤256-element coset (affine operand → `{(v−b)·a⁻¹ : v<bound}`), **refine-only** (only when it strictly shrinks an existing domain — `spec.bound < d0.size`, checked before building the coset), so the table is **bit-identical** to the ungated version and it never creates domains that would add enumeration work on VMs like OpenVM.

## Results (`profile`, min-of-3)
- SP1 keccak domainBatch **~3963 → ~830 ms**; SP1 rsp ~5× smaller.
- **OpenVM: no regression** (refine-only can't add work — coset built only when it shrinks).
- Effectiveness **byte-identical** on SP1 keccak (2665/313/2067), SP1 rsp, OpenVM keccak (2021/186/1748).

## Verification
`lake build` clean; `check-proof-integrity` passes (axioms propext/Classical.choice/Quot.sound only, no unused theorems). Audited surface untouched.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/DomainBatchRuntime.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/DomainBatchRuntime.lean
@@ -644,6 +644,82 @@ def denseForcedIdxV (cs : List (DenseConstraintPlan p)) (bis : List (DenseBusPla
     bisIdx := denseFreezeCovIndex (denseAnchorCovBuild (fun bi => bi.vars) bis),
     arrBis := bis.toArray }
 
+/-- Is `op` a recognized byte-relation selector for `spec`? Every such op guarantees both operands
+    are below `spec.bound` (`BusFacts.byteXorSpec_sound` / `byteBoolSound`). -/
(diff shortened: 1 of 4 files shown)


qwang98:
openvm-eth looks a bit suspicious with +9%, but I launched another run (breakdown above) and it shows that +9% is really a noise.

PR #235, opened 2026-07-28
perf: discriminate flagFold dedup buckets by first variable slot

qwang98:
Runtime-only, output-identical, zero proof change. Independent of #234 (touches only FlagFoldDrops.lean).

flagFold's pointwise-dedup (part C) is SP1's #1 pass (~27% of keccak apc_001, ~n^1.65). Bisection: the cost is the scan's `denseMsgEqCert`/`denseBoxAgree` calls (~16µs each), and SP1's byte/range checks share the opcode constant in slot 0 → collide into one giant bucket (1694 on keccak cycle 0; 70K certs, 0 drops). Fix: fold every constant payload slot's hash into the coarse key + index representatives by the FIRST variable-carrying slot (not slot 0) — both sound necessary conditions for `denseMsgEqCert`. Cycle-0 buckets 1694→5, certs 70K→1.7K.

Results: flagFold ~4.5–10% faster on SP1 keccak/rsp, OpenVM keccak ~−8% (no regression). Effectiveness byte-identical. ZERO proof change — `densePdDropSet` is a re-verified drop proposal (`densePdVerdictKeep_false` re-checks each drop against `densePdKeep`), so a tighter necessary-condition key cannot change output. `lake build` clean; `check-proof-integrity` passes. Residual cost (cycle-1 bucket / per-cert `denseBoxAgree`) noted as a follow-up in ideas.md.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/FlagFoldDrops.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/FlagFoldDrops.lean
@@ -152,28 +152,40 @@ structure DensePdEntry (p : ℕ) where
   bi : BusInteraction (DenseExpr p)
   sigs : Array (UInt64 × UInt64)
 
-/-- The coarse class key (bus id, constant multiplicity, payload length) and per-slot signatures
-    of one interaction — shared by both sweep paths. -/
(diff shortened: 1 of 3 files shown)


qwang98:
Closing as stale.

PR #238, opened 2026-07-28
perf: rewrite reencode's accepts in place instead of rebuilding the constraint list (reencode ~1.6x on sha256)

pacheco:
Every accept in `denseReencodeOut` rebuilt the whole constraint list — roughly 146 k cons cells on sha256, to drop 7 constraints and rewrite a few more. This makes an accept proportional to what it actually changes.

## What changes

The loop threads a `DenseReencodeWork` instead of a `DenseConstraintSystem`:

* a dropped constraint is **tombstoned as `.const 0` in place**, so positions stay stable and keep matching the bucket index (which names stable positions — a compacting drop would desynchronise it after the first accept);
* the rewrite rebuilds cons cells only as far as `maxPos`, the last position that *could* change, and **shares the untouched tail**;
* the booleanity constraints are accumulated and appended once at pass exit — appending them per accept would copy the whole spine and destroy that sharing.

`denseWorkView` is the system this represents.

## Why it needs no new correctness argument

`denseWorkOut_view` is the keystone: an accept on the working system is exactly `denseReencodeOut` on the view, **followed by dropping trivially-true constraints**. Both halves are already verified — `denseCheckReencode_sound` and `DensePassCorrect.denseFilterConstraintsEntailed` — composed with `.andThen`. Nothing new is claimed about the rewrite itself.

Two invariants, both threaded through the loop:

| invariant | what it gives | how it survives an accept |
|---|---|---|
| `DenseWorkPosOk` | `lastPos`/`lastFold` bound the positions mentioning each variable and the foldable ones; `denseWorkBounded_of_posOk` turns this into the bound that licenses sharing the tail | the splice only ever raises them, and raises them at every position it rewrites (`denseWorkSpliceCs_posOk`) |
| `DenseWorkBoolsOk` | every accumulated bool belongs to a minted bit | bools are appended together with their bits |

The second one is what makes deferring the bools sound. A deferred `b·(b−1)` must be inert for every later group, and every fact needed for that (not covered, fixed by the rewrite, not a placeholder) reduces to `b ∉ xs`. Rather than derive that from registry freshness, **two cheap runtime guards** supply it directly: no group variable and no freshly minted bit may be a bit minted earlier. Rejecting a candidate is always allowed, so the guards cost behaviour at worst, never soundness — the same device the pass already uses for `x ∉ bits` and the `powdrId?` check.

`denseCheckReencode_congr` lets the certificate read the working list directly. Materialising the view per candidate would allocate a whole spine on every candidate, which is exactly the cost this change removes.

## Relation to #231

This **supersedes** #231s per-accept fused gate on the constraint side, which is deleted (`denseGateCsGo`, `denseGateDeg`, `denseReencodeOutOk` and their lemmas) — the constraint side no longer walks the output at all. The bus side still uses `denseGateBisGo`, so #231s traversal survives there. `Scripts/unused-theorems.txt` loses its `dWithin`/`denseReencodeOutOk_snd` block: those existed to guarantee byte-identical output, which this change no longer claims (see below), and `List.all_congr_mem` is now genuinely reachable.

## Effectiveness

**Not byte-identical to `main`, deliberately.** The view drops `.const 0` constraints, including any already present in the pass input, which `trivialConstr` would otherwise remove a cycle later. So effectiveness can only *improve* on the constraint axis and cannot regress on it; variables and bus interactions are untouched by the change. I have not run the corpus sweep locally — **the benchmark decides**, which is why this is a draft.

## Measurements

The prototype this was derived from (same algorithm, behind `@[implemented_by]`, no proofs) measured on sha256, interleaved, one rep:

| variant | reencode | vs main |
|---|---:|---:|
| main | 26 185 ms | — |
| this design | **16 036 ms** | **1.63x** |

with byte-identical output at that stage, and 167 811 vs 177 629 ms pipeline total (1.058x). Expect the verified implementation to land near that; the certificate now reads the working list directly and the seed builds `lastPos` in one extra pass at entry, so it is not identical to the prototype. CI has the real numbers.

Caveat worth stating: reencode is ~15 % of sha256s pipeline and accepts nothing on several corpus APCs, so the corpus-level effect will be much smaller than 1.63x suggests.

`lake build` clean, no warnings. `Scripts/check-proof-integrity.sh` passes: three standard axioms only, no unused theorems.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/Proofs/Reencode.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/Proofs/Reencode.lean
@@ -1341,7 +1341,7 @@ theorem denseWorkStep_correct [Fact p.Prime] (b : DegreeBound)
     (reg : VarRegistry) (w : DenseReencodeWork p) (state : DenseReencodeCacheState p)
     (xs : List VarId) (freshBase : String) (bs : BusSemantics p)
     (hcov : (denseWorkView w).CoveredBy reg)
-    (hpos : DenseWorkPosOk w.lastPos w.lastFold w.cs)
+    (hpos : w.bounded = true → DenseWorkPosOk w.lastPos w.lastFold w.cs)
     (hbools : DenseWorkBoolsOk w.bitSet w.bools) :
     reg.Extends (denseWorkStep b reg w state xs freshBase).1
     ∧ (∀ i, (denseWorkStep b reg w state xs freshBase).1.isInput i = reg.isInput i)
@@ -1372,8 +1372,12 @@ theorem denseWorkStep_correct [Fact p.Prime] (b : DegreeBound)
     exact hD
   have hview : denseWorkView ro.1
       = (denseReencodeOut (denseWorkView w) xs bits hm).filterConstraints
-          (fun c => !denseIsZero c) :=
-    denseWorkOut_view b w xs bits hm hpos hbools hxsB'
+          (fun c => !denseIsZero c) := by
+    have h := denseWorkOut_view b (denseWorkEnsureBounded w) xs bits hm
+      (denseWorkEnsureBounded_posOk hpos)
+      (by rw [denseWorkEnsureBounded_bitSet, denseWorkEnsureBounded_bools]; exact hbools)
+      (by rw [denseWorkEnsureBounded_bitSet]; exact hxsB')
+    rwa [denseWorkEnsureBounded_view] at h
   have hpolyVars := denseCheckReencode_polyVars (denseWorkView w) xs bits hm hchkView
   have hxsInput : ∀ x ∈ xs, reg1.isInput x = true := fun x hx => by
     rw [hbii x]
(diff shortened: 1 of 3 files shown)


pacheco:
Reproduced the wasm-eth regression locally and fixed it in 9a8f015.

## Diagnosis

Every regressing APC has **zero accepts**, and each lost a flat 30–56 ms — a fixed per-cycle cost rather than anything scaling with work done:

| APC | accepts | main | eager seed |
|---|---:|---:|---:|
| wasm-eth/apc_036 | 0 | 694 ms | 750 (+56) |
| wasm-eth/apc_012 | 0 | 410 | 443 (+33) |
| keccak | 0 | 491 | 521 (+30) |

`denseWorkSeed` folded every constraint`s variable list into a `Std.HashMap` to build `lastPos`, once per pass application. reencode runs in every cleanup cycle, so an APC that never accepts paid that fold repeatedly for a bound it never consulted. Same trap #231`s comment flagged for `dWithin`, in a new field.

## Fix

`denseWorkEnsureBounded` computes the bound on the **accept path** instead. The seed is now `∅`/`0`/`bounded := false` — free. The threaded invariant weakens to `w.bounded = true → DenseWorkPosOk …`, satisfied vacuously by the seed, and `denseWorkOut` keeps its already-bounded contract so the splice and all its proofs are untouched.

## After

| APC | accepts | main | this branch |
|---|---:|---:|---:|
| wasm-eth/apc_036 | 0 | 700 ms | **703** |
| wasm-eth/apc_012 | 0 | 406 | **407** |
| keccak | 0 | 487 | **493** |
| sha256 | 64 | 26 493 | **21 979 (1.21x)** |

Zero-accept APCs are back to parity (+1 to +6 ms, at or under this machine`s run-to-run noise). sha256 also **improved from 1.13x to 1.21x** — the eager fold was being charged to sha256 once per cycle across all 11 cycles, so removing it helped there too. Pipeline total 177 669 → 173 024 ms (1.027x).

## Correction to the PR description

The 1.63x quoted in the body is the **prototype**`s number and I should not have presented it as the expectation. The verified implementation measures **1.21x** on sha256. The remaining gap is real work the prototype skipped: the certificate reads the working list through `denseCheckReencode_congr`, and `denseWorkNameCounts` recomputes a filter-and-length per accept where the prototype threaded a counter. Treat 1.21x as the number.

`lake build` clean, no warnings; `Scripts/check-proof-integrity.sh` passes.

PR #242, opened 2026-07-28
spec: require unconditional derivation coverage

georgwiese:
Require derivations to cover every output variable even without a real trace. Canonicalize returned derivations accordingly.

Checks: `lake build`, `lake build docs`, `Scripts/check-proof-integrity.sh`.

### DIFF
--- a/ApcOptimizer/Spec.lean
+++ b/ApcOptimizer/Spec.lean
@@ -293,21 +293,20 @@ def Circuit.isSoundReplacementOf (optimizedCircuit originalCircuit : Circuit p)
 -- ANCHOR_END: isSoundReplacementOf
 
 -- ANCHOR: isCompleteReplacementOf
-/-- Whether an optimized circuit is a complete replacement for an original one.
-    Assuming every input variable carries a powdr ID, `ds` contains only methods
-    for output variables and covers every output variable from the input,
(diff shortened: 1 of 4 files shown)


ApcOptimizer/Spec.lean
```diff
@@ -293,19 +293,23 @@ def Circuit.isSoundReplacementOf (optimizedCircuit originalCircuit : Circuit p)
...
+  -- variables, and the return derivations.
+  ds.cover originalCircuit.vars optimizedCircuit.vars ∧
+  -- For any admissible satisfying assignment of the original circuit, the
+  -- optimized circuit is also satisfied and admissible, with equivalent side
+  -- effects, under the assignment produced by witness generation.
+  ∀ assignment,
+    originalCircuit.admissible busSemantics assignment →
     originalCircuit.satisfies busSemantics assignment →
-    ds.cover originalCircuit.vars optimizedCircuit.vars ∧
-    (∀ derivation ∈ ds, derivation.1 ∈ optimizedCircuit.vars) ∧
```
georgwiese, line 308: This is the only semantic change in this PR: Since these two properties do not depend on `assignment`, they are moved out of the universal quantifier. This is a slightly stronger statement: They now also have to hold even if no admissible and satisfying assignment exists for the input circuit.

PR #253, opened 2026-07-29
perf(busPairCancel): decide the shield by a descending early-exit scan (0.33x on sha256)

pacheco:
## What

`busPairCancel`'s before-region *shield* was 65 % of the pass, and 92 % of those samples come from the shield scan alone. Its fold asks

```
ok(l) = ∀ k, P l_k ∨ ∃ k' > k, Q l_k'
```

which is decided entirely by the **topmost** visited live position `q` with `¬P q ∨ Q q`: `ok = P q` (`true` when there is none). So a descending walk with an early exit computes the *same boolean* while touching only the positions above the last provable receive.

- `DenseKeyIdx` also carries its buckets and `sym` as ascending arrays (`byKeyA` via `Std.HashMap.map`, `symA`); the lists stay as the proof-side spec.
- the mid test becomes a gated array walk (`denseLiveAllGated`), the shield a descending two-array early-exit walk (`denseShieldEarly`), with `denseShieldEarlySeg` for the symbolic-key fallback (no index needed there).
- both walks re-check the position window at every entry, so **no range search is needed**. A binary-searched variant was built and measured first: only 2 % faster on sha256 (6250 vs 6388 ms), not worth its extra correctness argument.

All under `Implementation/`; the audited surface is untouched.

## Why it is this big

Counted with a throwaway `dbg_trace` probe on `sha256/apc_001` (reverted): 14 848 candidates reach the region tests, 97 % are accepted, and the shield visits **28.6 M positions** (mean 1982). Two reasons: OpenVM's execution bridge has `addressFields = []`, so all 8006 bridge messages share **one** key bucket (4002 candidates × mean 4002 positions = 56 % of all shield work); and hot memory cells have 900–1500-entry buckets plus 1114 symbolic-address messages that every memory candidate visits.

The same probe ran the proposed walk next to the real scan: **28.6 M → 0.86 M positions (0.030×)**, verdict identical on **14 433 / 14 433** candidates — known before a line of proof was written.

## Proof

Two semantic lemmas replace the whole sparse-list layer, which is deleted (`denseMergeAsc`, `denseLiveAllSparse`, `denseShieldScanSparse` and their `_eq` lemmas):

- `denseLiveAllSegP_true_of` — per-position refutations give the segment `all`.
- `denseShieldScanSegP_true_aux` / `_of` — the shield fold holds when every live position is refuted except possibly below a witness position that is itself a live provable receive; one induction, carrying the fold's "a provable receive occurs later" flag leftwards.

`denseShieldEarly_sound` / `denseShieldEarlySeg_sound` / `denseLiveAllGated_sound` supply exactly those hypotheses (the per-array `Pairwise (· < ·)` from `denseKeyIdxBuild_sound` is what makes "every entry above the stop position was examined" true). Positions that are not index entries are refuted by the unchanged `denseMidRefuted_of_keyNe` / `_of_crossBus` + `densePreRefuted_of_midRefuted`.

## Numbers

Interleaved on one quiet 20-core box, median of 3 reps (single shot for sha256); busPairCancel, then the case total:

| case | busPairCancel | ratio | case total | ratio |
|---|---:|---:|---:|---:|
| `sha256/apc_001` (worst absolute) | 19276 → **6391 ms** | **0.332×** | 159266 → 146419 ms | **0.919×** |
| `wasm-eth/apc_028` (worst share) | 708 → **108 ms** | **0.153×** | 2345 → 1748 ms | 0.745× |
| `wasm-eth/apc_012` (near p95) | 2197 → 1186 ms | 0.540× | 16185 → 15297 ms | 0.945× |
| `keccak/apc_001` | 1143 → **469 ms** | **0.410×** | 12628 → 12020 ms | 0.952× |
| `openvm-eth/apc_100` (family tail) | 146 → 121 ms | 0.829× | 2074 → 2042 ms | 0.985× |

No other pass column moves outside run-to-run noise; `busPairCancelLate` is flat (250 → 239 ms).

## Verification

- **Output byte-identical** (`opt-export` + `cmp`): `sha256/apc_001` (5.76 MB export), OpenVM `keccak/apc_001`, `openvm-eth/apc_006`, `openvm-eth/apc_100`, `wasm-eth/apc_012`, `wasm-eth/apc_028`, SP1 `keccak/apc_001`.
- keccak `run` counts identical: 2021 / 186 / 1748.
- `lake build` clean, no warnings; `Scripts/check-proof-integrity.sh` passes (axioms `propext` / `Classical.choice` / `Quot.sound`, no `sorry`, no unused theorems).

Effectiveness should be unchanged everywhere — the transformation is value-preserving — but the CI matrix is the arbiter.

## Residual (for the next pass at this)

The remaining 6.4 s on sha256: the surviving 0.86 M certificate evaluations (`denseAddrNonzeroNeqP` is allocation- and `ZMod`-dictionary-bound — `denseIsZeroLin` derives the whole `CommRing` chain *before* testing `terms.isEmpty`), the per-invocation index builds (5.4 % of the old pass), the per-drop `alive` array copy (2.1 %, ~8 GB of memcpy over 14 418 drops) and `denseInteractionBound`'s per-variable pattern rebuild. Written up in `agent-docs/ideas.md` R8 and log entry 153.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/BusPairCancelKeyIdx.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/BusPairCancelKeyIdx.lean
@@ -10,9 +10,10 @@ the pass. For a candidate whose address slots are all constants, a message can o
 region tests if it is on the same bus and either shares the candidate's constant address key or
 has a non-constant key: every other position is refuted by the bus-id or constant-disequality arm
 and contributes the identity to the scan fold. `DenseKeyIdx` buckets each bus's positions by
-constant address key (plus a `sym` list for non-constant keys), so the sparse scans below visit
-only the same-key and symbolic positions. `Proofs/BusPairCancelKeyIdx.lean` proves the builder
-sound and packages the sparse results back into the full-region forms (`denseRegionTests`). -/
+constant address key (plus a `sym` list for non-constant keys, and both as arrays), so the scans
+below visit only the same-key and symbolic positions — and the shield stops at the first position
+that decides it. `Proofs/BusPairCancelKeyIdx.lean` proves the builder sound and packages the scan
+results back into the full-region forms (`denseRegionTests`). -/
 
 namespace ApcOptimizer.Dense
 
@@ -35,6 +36,10 @@ def denseKeyHash (k : List (ZMod p)) : UInt64 :=
 structure DenseKeyIdx (p : ℕ) where
   byKey : Std.HashMap UInt64 (List Nat)
   sym : List Nat
+  /-- `byKey`/`sym` as ascending arrays: the runtime walks these (`denseLiveAllGated`,
+      `denseShieldEarly`) instead of materializing a filtered list per candidate. -/
+  byKeyA : Std.HashMap UInt64 (Array Nat)
+  symA : Array Nat
 
 /-- Insert one position (front of its bucket; the builder folds positions descending). -/
 def denseKeyIdxAdd (shape : MemoryBusShape) (busId : Nat)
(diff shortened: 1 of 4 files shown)


pacheco:
## The runtime row in the effectiveness matrix is the parallel-harness artifact, not a regression

The matrix reported wasm-eth +20 %, SP1 rsp +67 %, openvm-eth +5 % (while keccak −4 %, sha256 −6 %, SP1 keccak −7 %). That row comes from `Benchmarks/benchmark.py`, which runs cases with `--jobs = cpu_count`; its own docstring calls the runtime "rough", and `agent-docs/ideas.md` records this row swinging **+51 % → +15 % on identical code** for openvm-eth.

Two independent checks say there is no regression.

**1. CI's own serial harness** (`Runtime Bench`, run 30476080811, same runner, target vs baseline) on openvm-eth, 100 cases:

| | target | baseline | Δ |
|---|---|---|---|
| total | 101.2 s | 102.5 s | **0.99×** |
| median case | 84 ms | 85 ms | 0.99× |
| `busPairCancel` | 1399 ms | 1689 ms | **0.83×** |

Every other pass is 1.00× ± noise.

**2. Local serial re-run of the three flagged sets**, interleaved per case (baseline then this branch, one process at a time, 20-core box idle):

| set | cases | wall (main → PR) | Δ | `busPairCancel` | Δ |
|---|---:|---|---:|---|---:|
| SP1/rsp | 100 | 11409 → 11202 ms | **0.982×** | 938 → 715 ms | 0.762× |
| openvm-eth | 100 | 36044 → 35691 ms | **0.990×** | 1794 → 1469 ms | 0.819× |
| wasm-eth | 100 | 71514 → 66252 ms | **0.926×** | 10017 → 5079 ms | 0.507× |
| **all three** | 300 | 118967 → 113145 ms | **0.951×** | 12749 → 7263 ms | **0.570×** |

Per-case ratios over the 153 cases ≥ 50 ms: median 0.992, mean 0.988, range 0.736–1.139. The outliers in both directions are 8–80 ms cases where `busPairCancel` itself is 0–5 ms, i.e. ±5 ms of timer noise — on SP1 rsp, 28 cases are >5 % slower and 30 are >5 % faster, and the 15 cases ≥ 200 ms come out at 0.980×. **Every one of the 23 cases where the pass costs ≥ 50 ms on main improves** (0.16×–0.97×).

Full-set timings for the other three sets, single-process: keccak 12628 → 12020 ms (pass 0.410×), sha256 159266 → 146419 ms (pass 0.332×), SP1 keccak 6556 → 6149 ms (pass 901 → 481 ms, 0.53×).

**Memory was the one plausible mechanism for a parallel-only regression** (the index is now stored as both lists and arrays), so I measured peak RSS: wasm-eth apc_012 241.7 → 228.1 MB, openvm-eth apc_005 111.8 → 112.4 MB. No growth worth a scheduling effect.

Effectiveness remains exact on all six sets, per-case sizes identical.

pacheco:
## SP1 rsp, serial, on a CI runner: 0.95× total, `busPairCancel` 0.77×

The parallel effectiveness row read **+67 %** for this set. `Runtime Bench` could not reach SP1 sets (its `benchmark` input feeds a script that only resolved OpenVM subdirectories), so I made the selector accept `<vm>:<set>` — #254 — and dispatched it on a scratch branch carrying this PR's commits plus that selector, against `main` on the same runner: [run 30478074537](https://github.com/powdr-labs/apc-optimizer/actions/runs/30478074537).

**SP1 rsp, 100 cases, serial, target vs baseline (same runner):**

| | target | baseline | Δ |
|---|---|---|---|
| total | 7035 ms | 7388 ms | **0.95×** |
| mean | 70 ms | 73 ms | 0.96× |
| median | 26 ms | 28 ms | — |
| max | 562 ms | 600 ms | 0.94× |

| pass | target | baseline | Δ |
|---|---|---|---|
| `busPairCancel` | 447 ms | 584 ms | **0.77×** |
| `busPairCancelLate` | 53 ms | 67 ms | 0.79× |
| domainBatch | 1229 ms | 1268 ms | 0.97× |
| busUnify | 769 ms | 795 ms | 0.97× |
| flagFold | 610 ms | 602 ms | 1.01× |
| carryBranch | 516 ms | 511 ms | 1.01× |

The five slowest rsp cases come out 0.86×–0.97×; nothing regresses. This matches my local serial run of the same set (0.982× wall, 0.762× pass) and CI's serial openvm-eth run (0.99× / 0.83×).

So all three serial measurements — two on CI runners, one local — agree, and the only numbers showing a regression are the ones `benchmark.py` produces with `--jobs = cpu_count`.

**Noise floor for that harness**, measured the same way on a branch whose optimizer is identical to main ([run 30480176914](https://github.com/powdr-labs/apc-optimizer/actions/runs/30480176914)): SP1 rsp total 9747 vs 9755 ms = **1.00×**, `busPairCancel` 1.02×. So the 0.95× / 0.77× above is signal, and +67% is not this harness's variance.

PR #254, opened 2026-07-29
tooling: accept a <vm>:<set> benchmark token

pacheco:
`Runtime Bench` forwards only its `benchmark` input to `runtime_bench.py`, never `--vm`, so SP1 sets could not be dispatched serially. (`--vm sp1` works locally; `-f benchmark=../SP1/rsp` finds the directory but then runs the cases without the VM token and crashes.)

The positional argument now also accepts `<vm>:<set>`, the form `lean_action_ci.yml` already uses:

```sh
gh workflow run "Runtime Bench" -f pr=253 -f benchmark=sp1:rsp
```

Bare names and `--vm` behave as before, the two conflicting is an error, headings name the VM for non-OpenVM sets, unknown sets list the corpus, and `--compare` still reads pre-change JSONs.

Tried on CI: [run 30478074537](https://github.com/powdr-labs/apc-optimizer/actions/runs/30478074537) benched SP1 rsp serially on both sides (total 7035 vs 7388 ms) — the measurement #253 needed.

Alternative shape: a `vm` input on the workflow, which is cleaner but needs a `.github/workflows/` push I'm not scoped for; that patch is 8 lines if you prefer it.

### DIFF
--- a/Benchmarks/runtime_bench.py
+++ b/Benchmarks/runtime_bench.py
@@ -11,7 +11,8 @@ per pass across the whole set. Cases run *serially* so timings don't fight for c
 This measures runtime only; effectiveness is benchmark.py's job.
 
     Benchmarks/runtime_bench.py                 # all openvm-eth cases
-    Benchmarks/runtime_bench.py --vm sp1        # all rsp (SP1) cases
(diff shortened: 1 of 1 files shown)


pacheco:
@codex review

PR #257, opened 2026-07-30
perf(derivations): index the output-derivation pruning (3.7x on the openvm-eth family)

georgwiese:
## What regressed

Timing the Lean optimizer against powdr's native Rust optimizer over 20,115 real APC input circuits (9 benchmarks, guest suite + openvm-eth block 24171377) showed total Lean time going **7,202s → 10,832s (+50%)** between `f38461c` and `0a5c983`, while the Rust side over the same circuits stayed flat (−2%).

Bisecting over the 18 commits in that range, on five openvm-eth APCs of 3.2k–4.3k variables that regressed hardest, run alone (Lean optimizer time only, sum over the five):

| rev | | |
|---|---|---|
| `f38461c` | 6.73s | good |
| `e7292cb` (#240) | 6.95s | good |
| `21102c2` (#244) | 6.39s | good |
| **`74515b5` (#242)** | **24.05s** | **bad** |
| `e180ffe` (#246) | 23.75s | bad |
| `0a5c983` (main) | 22.51s | bad |

Its parent is clean and nothing after it recovers, so #242 accounts for the whole +50%. The spec hoist in that PR is not the problem — requiring `Derivations.cover` without a trace is a genuine strengthening, and it's a good change. What costs the time is *how* the implementation discharges it.

## Why it costs time

To make `cover` structurally true, `forOutput` builds a total, sanitized derivation map over the no-ID output variables. It did that with three list scans, each superlinear because **both `Circuit.vars` lists count variable occurrences, not distinct variables**:

- `Derivations.methodFor` walks *all* of `ds` per looked-up variable — it recurses before testing the head (that's what makes the last binding win), so there is no early exit;
- `List.eraseDups` tests each occurrence against the kept-so-far list (`eraseDupsBy.loop`);
- the `∀ x ∈ cm.vars, x ∈ inputVars` safety test rescans `inputVars` per referenced variable.

`Variable` equality is `String` equality, so each comparison in those scans is a string compare. (The pre-#242 line was already quadratic — `derivs.filter (·.1 ∈ out.vars)` is `|ds|` membership scans over an occurrence list — which is why this ends up slightly faster than pre-#242 as well.)

## The fix

Serve all three from indexes, and keep the list-shaped definition as a proven-equal reading:

- `ds` keyed by variable via `Std.HashMap.ofList`, whose `insertMany` lookup lemma keeps the **last** binding for a duplicated key — exactly `methodFor`'s rule (`getElem?_ofList`, via `methodFor_eq_findSomeRev?`);
- the input variables as a `Std.HashSet` (`Std.HashSet.contains_ofList`);
- the existing, already-proven `HashedDedup.hashedEraseDups` for the dedup.

`forOutput_eq` rewrites the indexed definition back to the previous list-shaped one, so **the output list is bit-for-bit unchanged** and the correctness proof keeps its previous shape (two `rw` sites). No change to the audited surface; no new `csimp`, `sorry`, or trusted implementation.

`lake build` is clean (no warnings) and `Scripts/check-proof-integrity.sh` passes: correctness theorems still rest only on `propext`/`Classical.choice`/`Quot.sound`, and no theorem is orphaned.

## Measurements

Isolated on a 48-core box (each circuit run alone, so these are lower than the loaded full-suite numbers), Lean optimizer time only:

| circuit | `f38461c` (pre-#242) | `0a5c983` (main) | this PR |
|---|---|---|---|
| five 3.2k–4.3k-var openvm-eth APCs (sum) | 8.17s | 23.69s | **7.31s** |
| one 29,340-var openvm-eth APC | 21.91s | 167.26s | **18.34s** |

So **3.2× on the small five and 9.1× on the 29k-variable APC** versus main, landing slightly ahead of pre-#242 (the small-circuit numbers carry ~20% run-to-run variance on this box; the 29k figure is well outside it). Effectiveness is unchanged by construction — the returned derivation list is proven identical.

https://claude.ai/code/session_01HKR42FVxb2GLyEesKGgyvK

### DIFF
--- a/ApcOptimizer/Implementation/Optimizer.lean
+++ b/ApcOptimizer/Implementation/Optimizer.lean
@@ -23,6 +23,7 @@ import ApcOptimizer.Implementation.OptimizerPasses.Proofs.DegenRange
 import ApcOptimizer.Implementation.OptimizerPasses.IdentitySubst
 import ApcOptimizer.Implementation.OptimizerPasses.Proofs.IdentitySubst
 import ApcOptimizer.Implementation.OptimizerPasses.DenseUmbrella
+import ApcOptimizer.Implementation.OptimizerPasses.HashedDedup
 
 set_option autoImplicit false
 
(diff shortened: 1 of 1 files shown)


georgwiese:
Cool, strictly better than pre-#242 numbers!

PR #258, opened 2026-07-30
ci: bench baseline and PR on the same runner for each comparison

pacheco:
The effectiveness matrix crossed `side: [pr, main]` with the benchmark group, so each set's two sides landed on different runner instances. Circuit sizes are deterministic and don't care, but benchmark.py also reports each side's wall time and `report` assembles those into the runtime row -- comparing two machines. That row has already misled us: a change a same-runner bench puts at 0.86x on sha256 was reported here as -2%, and the same run over-reported openvm-eth at -30% against a measured 0.90x. Two hypotheses were investigated before the measurement itself was suspected.

Now one cell per group, benching both sides itself: PR binary then latest-main binary, back to back on that cell's runner. Groups still fan out across runners, which is where the parallelism belongs; only the baseline/PR pair is pinned together.

- PR side runs first, so the second (main) side gets the warm page cache. A reported improvement is then a lower bound rather than an overstatement.
- Each cell uploads `eff-pr-<key>` and `eff-main-<key>`, exactly the artifact names the per-side cells used to produce, so `report` merges them unchanged.
- Per-group timeouts, since a cell now runs its sets twice. Measured on a recent run: the sha256 cell is ~170 s per side and the others under a minute, so 30/30/45 min is headroom for the stress case growing rather than a current requirement. The old comment claimed "~20 min for one optimizer run", stale by ~7x.
- Output paths are absolute: benchmark.py chdirs to the repo, so relative `--json` worked only because the step's cwd happened to be the workspace.

The runtime row is now comparable, so the sticky comment no longer calls it report-only -- but it stays coarse, since benchmark.py runs cases in parallel. Per-pass detail remains the on-demand Runtime Bench, which already benched both sides on one runner.

### DIFF
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -32,73 +32,119 @@ jobs:
           retention-days: 30
 
   # Effectiveness over the full OpenVM (openvm-eth, wasm-eth) and SP1 (rsp) sets plus the keccak and
-  # sha256 stress sets. The matrix crosses side (this PR vs the latest main build) with a benchmark
-  # group, so each cell is its own runner: circuit sizes are deterministic, so the sides needn't
-  # share a runner, and splitting the 100-case wasm-eth set into its own group keeps it off the
-  # openvm-eth critical path. `report` merges every cell. benchmark.py also captures a rough total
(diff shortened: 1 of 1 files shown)


pacheco:
@codex review

PR #281, opened 2026-08-04
perf(guardDegree): thread a degree certificate and skip the items the output shares with its input (guard 0.31-0.37x, total 0.93-0.95x)

pacheco:
`guardDegree` is not a pass but the wrapper `guardAll` puts on **every** entry of `preludePasses` /
`cleanupPasses` / `codaPasses`, so `withinDegreeB` — a whole-system AST walk — runs **200–290 times
per run**. Against a build with the guard deleted it is **8–10 % of total wall** (it grew from the
6–7 % measured earlier: #276–#280 cut the passes, not the guard).

A probe classifying every AST node the guard visits says most of that work re-decides degrees it has
already accepted. Of 7.79 M / 64.9 M / 36.1 M nodes (openvm-eth `apc_006` / wasm-eth `apc_012` /
keccak `apc_001`): **35–45 % sit in a remaining list that *is* the input's remaining list**, another
**23–25 % in an item pointer-identical to the aligned input item**, and only 30–39 % are genuinely
rebuilt.

## The change

1. **A degree certificate threaded through the guarded stage lists.**
   `DenseDegCert b d := Option (PLift (d.withinDegreeB b = true))` — the proof is erased, so a
   certificate is a constructor tag at runtime. `guardAll` now returns `DenseGuardedPassW p b`
   (system + certificate in, result + certificate out) and *produces* `some` whenever its own check
   passes, so the certificate establishes itself on the first invocation of a stage — the optimizer's
   input circuit is still not assumed within bound. Stage-list entries stay plain
   `DenseVerifiedPassW`, so AGENTS.md's one-line pass-adding contract is unchanged;
   `DenseGuardedPassW.toPass` feeds `none` and hands the chain back to `denseIterateToFixpoint` and
   `pipeline`.
2. **A sharing-aware check** (`DenseConstraintSystem.degOkFrom`, `Measure.lean`): walk the output's
   item list in lockstep with the input's, using core Lean's `withPtrEq` at two granularities — a
   pointer-identical *remaining list* ends the scan, a pointer-identical *item* skips its AST walk.
   `withPtrEq a b k h` is definitionally `k ()`, so these are hints and nothing more: `degOkFrom_eq`
   proves the result equals `withinDegreeB`, and the compiler emits a bare `lean_ptr_addr` compare
   with no closure allocation. Each definition returns its own specification
   (`{ r : Bool // r = ol.all ok }`) because `withPtrEq`'s `h` obligation has to be discharged while
   the recursion is elaborated; a one-runtime-field `Subtype` is that field. Two monomorphic copies
   (constraints, interactions) keep the per-item test a direct call, which also removes the boxed
   closure `List.all` allocated per item.
3. `attribute [local irreducible] densePipeline` after `densePipeline_respectsDeg` — the extra
   definitional layer pushed `optimizerWithBusFacts_correct`'s `change`s past the heartbeat limit.
   Sealing the pipeline where it is used only through its `DensePassResult` fields takes that file
   from 8.7 s to 1.3 s, faster than before this change.

`Main.lean`'s `profile` threads the certificate the same way the pipeline does (dropped at the three
stage boundaries and at each fixpoint iteration), so the reported per-pass times stay faithful.

## Numbers

20-core box, serial, interleaved, 2–3 reps; sha256 one shot per rep (first discarded as cold):

| APC | main 75bf56f | this branch | total | guard |
|---|---:|---:|---:|---:|
| openvm-eth `apc_006` | 315 ms | 293 | **0.930×** | 30 → 8 ms |
| wasm-eth `apc_012` | 2653 | 2510 | **0.946×** | 228 → 85 |
| keccak `apc_001` | 2203 | 2075 | **0.942×** | 185 → 57 |
| sha256 `apc_001` | 24055 | 22563 | **0.938×** | — |

That recovers 63–73 % of the guard's cost. What shrinks per pass is the entries that share their
items (`carryBranch`, `bytePack`, `busPairCancel`, `tautoBus`, `disconnected`, `rootPairUnify`); what
still pays is the entries that rebuild everything (`normalize1/2`, `constFold0/1/2`, `gauss`,
`domainBatch`). A pass that shares nothing pays two pointer compares per item on top of the walk it
would have done anyway.

## Verification

- `opt-export` **byte-identical** on 8 cases: openvm-eth `apc_006` / `apc_100`, wasm-eth `apc_012` /
  `apc_063`, OpenVM keccak `apc_001`, SP1 keccak `apc_001`, SP1 rsp `apc_001` / `apc_002`. This is a
  pure runtime change, so effectiveness cannot move.
- `lake build` clean, no warnings; `Scripts/check-proof-integrity.sh` passes (no `sorry`; the
  correctness theorems rest on `propext`, `Classical.choice`, `Quot.sound`).

## Measured dead ends (recorded in `agent-docs/`)

- **Sub-expression lockstep down `add` spines** — sound (every subterm of a certified expression is
  certified), worth **0.01 % of nodes**: the rebuild passes materialize fresh trees all the way down.
- **An unboxed `UInt64` honest walk** (pure register recursion in the C) — measured **flat**: that
  walk is memory-bound on the AST. A first version capped at `2^32` was 5 % *worse* because Lean
  emits `lean_cstr_to_nat` per call for a literal above the scalar range.
- **Per-pass degree monotonicity (no check at all)** — sized with `sorry` certificates: the easy set
  (`constFold*` plus the drop passes) is **~1 % of the run**; it only pays with `normalize` and
  `gauss`, which need real per-pass degree theorems.
- **Certificate through `denseIterateToFixpointFrom`** — ~0.5 % of keccak, against a second copy of
  the fixpoint definition and its induction.
- **Resynchronizing the lockstep cursor after a drop or prepend** — impossible in the safe fragment:
  `withPtrEq`'s outcome cannot steer control flow.

Draft: opening for the CI benchmark and effectiveness run.

### DIFF
--- a/ApcOptimizer/Implementation/Optimizer.lean
+++ b/ApcOptimizer/Implementation/Optimizer.lean
@@ -42,21 +42,21 @@ correctness theorems in `ApcOptimizer/Optimizer.lean` are projections of `optimi
 /-- Wrap every pass of a stage list in the degree guard, so degree safety holds uniformly
     (`guardAll_chain_respectsDeg`) with zero per-pass proof burden. -/
 def guardAll (b : DegreeBound) (l : List (String × DenseVerifiedPassW p)) :
-    List (String × DenseVerifiedPassW p) :=
+    List (String × DenseGuardedPassW p b) :=
   l.map (fun (n, f) => (n, f.guardDegree b))
 
 /-- Stage 1 of 3 (`preludePasses`, `cleanupPasses`, `codaPasses`): run once to canonicalize the
     freshly-parsed system. The three lists are the single source of truth for the pass sequence —
     `pipeline` folds them and the `profile` CLI (`Main.lean`) times the same lists, so they cannot
     drift. `String` labels name passes in the profiler only. -/
-def preludePasses (b : DegreeBound) : List (String × DenseVerifiedPassW p) :=
+def preludePasses (b : DegreeBound) : List (String × DenseGuardedPassW p b) :=
   guardAll b [ ("constFold0", denseConstantFoldPass) ]
 
 /-- Stage 2 of 3 (see `preludePasses`): the cleanup schedule, iterated to a fixpoint
     (`denseIterateToFixpoint`, no budget). Each entry's optimization is documented at its own
     definition. To add a pass, append one `(name, pass)` entry here (AGENTS.md →
     "Adding an optimization"). -/
-def cleanupPasses (b : DegreeBound) : List (String × DenseVerifiedPassW p) :=
+def cleanupPasses (b : DegreeBound) : List (String × DenseGuardedPassW p b) :=
   -- One primality decision per run, threaded to every prime-gated pass (each reads the `Bool` in
   -- O(1) instead of re-running `decide (Nat.Prime p)`).
   let pw := PrimeWitness.of p
@@ -92,7 +92,7 @@ def cleanupPasses (b : DegreeBound) : List (String × DenseVerifiedPassW p) :=
 /-- Stage 3 of 3 (see `preludePasses`): run once after the cleanup fixpoint. Order matters — the
     inline notes below flag the non-obvious sequencing; each pass's own definition documents what it
     does. -/
-def codaPasses (b : DegreeBound) : List (String × DenseVerifiedPassW p) :=
+def codaPasses (b : DegreeBound) : List (String × DenseGuardedPassW p b) :=
   let pw := PrimeWitness.of p
   guardAll b
   [ ("busPairCancelLate", denseBusPairCancelPass pw true),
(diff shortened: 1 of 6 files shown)


pacheco:
@codex review

pacheco:
Closing in favour of #290, which now carries this PR's two load-bearing walk refinements — the remaining-list `withPtrEq` shortcut and the monomorphized per-item walkers — with credit, in a certificate-free architecture (the `outOk || !inOk` per-item test discharges `withPtrEq`'s obligation by Boolean reasoning, so no `DenseGuardedPassW` threading and no stage-list type change; `Pass.lean` only).

The deciding three-way A/B (both rebased onto the same main, outputs bit-identical): the hybrid matches or beats this PR on sha256 (0.940× vs 0.948–0.956×), apc_036 and keccak; this PR keeps a consistent ~4-point edge on wasm apc_012 only (0.922× vs 0.962×) — unexplained, and worth revisiting if apc_012-shaped cases ever dominate. The certificate architecture itself measured as neutral on aggregate; if a real tenant for threaded per-pass state lands later (F3/F4-style), this PR is the reference design for the plumbing.

PR #282, opened 2026-08-04
perf(tupleRange): the pack scan gives up at the first candidate without a partner (0.08x on the pass)

pacheco:
`tupleRange` was **852 ms of sha256 `apc_001` and packed nothing there** — 2.7 % of the local corpus
in a coda pass that had never been profiled, and the only genuinely quadratic loop left in the top 20.

`denseFindTuplePack` walked the interaction list and, at every position `denseMatchByteSingle`
recognized, called `denseFindRangePartner` over the whole remaining list — then, **on failure,
recursed to the next position and rescanned the remainder again** (mirrored for the
range-check/`denseFindBytePartner` branch). sha256 is full of byte self-checks and has no exact-width
range partner for the declared tuple bus, so every one of them paid a full-list scan to find nothing.
Stubbing `denseMatchByteSingle` to `none`, so no partner scan can start, took the pass from 852 ms to
3 ms — which located the whole cost before anything was written.

## The change

The two "partner search failed" recursions become `none`. The scan is then one left-to-right pass:
positions matching neither recognizer are skipped into `revPre`, and at the *first* position matching
either one the suffix is scanned once for its complement.

## Why the output is unchanged

Let `p` be the first position matching either recognizer. Positions before `p` are skipped the same
way by both versions, and a hit at `p` produces the same `pre`/`mid`/`post` split. If `p`'s complement
is absent from `(p, ∞)`, no later position can succeed either: each branch's guard and its partner
scan test the **same** recognizer at the same `bs`/`facts`/`s2`, so a later position either enters the
branch whose complement is missing from that same suffix, or is gated on the recognizer `p` matched
and scans a sub-suffix of `(p, ∞)`. Stating it over the recognizers rather than a "kind" partition is
what makes it hold for an interaction matching both, where the byte branch wins the tie.

So the pass makes the same pack decisions in the same order, and `denseDrainTuplePacks` above it is
untouched. **No new proof obligation**: both dropped branches are unreachable in
`denseFindTuplePack_split`, whose proof loses those two cases (−14 lines).

## Numbers

20-core box, serial, interleaved, 3 reps, medians (sha256 2 reps):

| case | pass | total | |
|---|---:|---:|---:|
| sha256 `apc_001` | **852 → 3 ms** | 23 336 → 22 333 | **0.957×** |
| wasm-eth `apc_028` | 85 → 1 | 443 → 366 | **0.826×** |
| wasm-eth `apc_006` | 93 → 24 | 1627 → 1569 | 0.964× |
| wasm-eth `apc_012` | 73 → 17 | 2653 → 2555 | 0.963× |
| wasm-eth `apc_037` | 43 → 8 | — | — |
| openvm-eth `apc_100` | 5 → 2 | 388 → 385 | 0.992× |
| keccak `apc_001` | 7 → 1 | 2243 → 2229 | 0.994× |

**Whole local corpus, 303 APCs, interleaved: `tupleRange` 1283 → 102 ms (0.080×), wall
50.5 → 49.2 s (0.975×), and no case where the pass is slower.**

## Verification

- Final `vars / constraints / bus interactions` **identical on all 303 corpus APCs**.
- `opt-export` byte-identical on wasm-eth `apc_006` / `apc_012` / `apc_037` / `apc_028`, openvm-eth
  `apc_100`, keccak `apc_001` — the first four being cases where the pass fires.
- `lake build` clean, no warnings, no `sorry`; `Scripts/check-proof-integrity.sh` passes, correctness
  theorems on the three standard axioms, no orphaned theorem.

## Residual (recorded in `ideas.md`)

102 ms corpus-wide, 61 of it on the six cases where the pass *fires*: `denseDrainTuplePacks` still
restarts at position 0 after every pack and recomputes `maxId` plus `denseTupleBusCandidates` each
step, so a firing drain stays `O(packs × B)`. A position cursor fixes it. `denseMatchByteSingle` also
runs `denseByteShape?` per position with no `@[csimp]` runtime twin, which is what the surviving
single pass costs.

The docs commit additionally records the cross-cutting measurements from the profiling session that
found this: the fresh whole-corpus attribution, **R15** (42 % of pass wall is invocations that return
their input, 37 % net of the guard, with the per-pass table and the four mechanisms behind it), and
**R5**'s corpus-wide guard cost (0.914×, 10 % of pass time).

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/Proofs/TupleRange.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/Proofs/TupleRange.lean
@@ -236,12 +236,7 @@ theorem denseFindTuplePack_split (bs : BusSemantics p) (facts : BusFacts p bs) (
       obtain ⟨specA, xA⟩ := sx
       simp only [hba] at h
       cases hrp : denseFindRangePartner bs facts s2 [] rest with
-      | none =>
-        simp only [hrp] at h
-        obtain ⟨D₁, D₂, spec, bw, heq, hcase⟩ := ih (a :: revPre) pre x mid y post h
-        exact ⟨D₁, D₂, spec, bw,
-          by simpa only [List.reverse_cons, List.append_assoc, List.singleton_append] using heq,
-          hcase⟩
+      | none => simp only [hrp, reduceCtorEq] at h
       | some res =>
         obtain ⟨mid', yB, bwB, post'⟩ := res
         simp only [hrp, Option.some.injEq, Prod.mk.injEq] at h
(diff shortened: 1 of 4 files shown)


pacheco:
@codex review

PR #288, opened 2026-08-05
Order-free memory-bus admissibility: multiset discipline + TS_BOUND rely

1arie1:
## Summary

This PR replaces the positional memory-bus rely (`admissibleMemoryBus`, which trusts that the input's bus-interaction list is *ordered by time* and asserts payload copying between list-adjacent same-address pairs) with an **order-free discipline**: properties of the *multiset* of evaluated messages, provably invariant under any reordering of the interaction list.

The new audited rely, per VM, is four short conjuncts:

1. **`admissibleMemoryBusM`** — at every *evaluated* address, the receives' payload multiset exceeds the sends' by at most one element (the entry receive). This is the multiset shadow of two system-level facts: bus balance (a received record is a sent record, consumed injectively — the defining property of the bus argument) and window atomicity (per address, at most one record enters the block from outside).
2. **`tsBounded`** (OpenVM only) — every message on a timestamp-declared bus (memory: payload slot 6; execution bridge: slot 1) has timestamp value < 2^29, i.e. OpenVM's global timestamp invariant. Per-message, hence trivially preserved by dropping messages and reordering.
3. **`entryKeyed`** (OpenVM only, and only when the caller supplies the block's entry pc) — the record entering the block from outside on the execution bridge carries that pc in payload slot 0. See *The one new assumption* below.
4. **`x0ReturnsZero`** — unchanged.

Per-VM sanity theorems (`openVmAdmissible_perm`, `sp1Admissible_perm`) prove the *entire* rely of each VM is invariant under `List.Perm` — no ordering assumption can hide anywhere.

## Why

- The old rely couples memory-pass completeness to an unverifiable frontend property ("interactions are listed in chronological order"). The new rely is stated against facts an auditor can check independently of emission order, and is strictly closer to definitional properties of the bus argument.
- The optimizer's output becomes order-robust: correctness no longer depends on the emission order of the input, and passes may in principle reorder interactions.
- The old positional discipline is *recovered as a theorem* on the canonical access order (`Implementation/MemoryBusMultiset.lean`, `interleaveAccesses_admissibleMemoryBus_of_M`) — the trust direction is inverted.

## The one new assumption (ENTRY_KEY)

Order-freeness costs one thing the list order paid for silently: **which record entered the block from outside**. The multiset cannot say. Take a 3-instruction loop body at pcs 100/104/108 whose last instruction branches back to 100, and compare the intended filling of the row with a *rotated* one — entered at instruction 1, wrapping through the block's exit back to instruction 0:

| | receives | sends | | receives | sends |
|---|---|---|---|---|---|
| instr 0 | (100, 10) | (104, 13) | | (100, 16) | (104, 19) |
| instr 1 | (104, 13) | (108, 16) | | (104, 10) | (108, 13) |
| instr 2 | (108, 16) | (100, 19) | | (108, 13) | (100, 16) |

Left: excess = {(100,10)}, card 1. Right: excess = {(104,10)}, card 1 — **equally admissible**, every timestamp increasing along the (rotated) chain, every per-instruction constraint satisfied. But the chain equality `ts₁ = ts₀ + 3` is false on the right. So it is *not entailed*, and emitting it would break completeness rather than merely lack a proof. Rotation is a property of the assignment, not of the interaction list, so no permutation-invariant predicate over messages excludes it — and it is not a cycle, so timestamp arithmetic does not either.

`entryKeyed` supplies exactly the missing datum: every payload in the excess multiset carries a declared key in a declared slot. For OpenVM's execution bridge that is the block's entry pc (powdr's `block.blocks[0].start_pc`, or a top-level `entry_pc` for callers that serialize only `{machine, bus_map, next_free_id}` — as `autoprecompiles/src/optimizer/lean.rs` currently does). Notes for review:

- **Opt-in.** `entryPc = none` ⇒ `memEntryKeyOf = none` ⇒ the conjunct is vacuous and the pass that consumes it emits nothing. Nothing is assumed unless a caller declares the pc.
- **Execution bridge only.** The memory bus declares no entry key; SP1 declares none at all.
- **Completeness-side only.** `admissible` does not appear in `isSoundReplacementOf`, so a wrong pc can cost a rejected real trace, never an accepted bad one.
- **Checkable.** The declared pc should be the constant the input's own constraints pin the first instruction's `from_state__pc` to (true in all 202 benchmark exports). README's *Assumptions (OpenVM)* section states the obligation.

This is a weaker assumption than the one it replaces: list-order trust named the entry (the first receive) *and* settled every "what lies between these two messages" question; ENTRY_KEY buys back only the naming, for one bus.

## What must be reviewed (audited surface)

`Spec.lean` is untouched. The review is ~170 changed lines across four files:

- `ApcOptimizer/MemoryBus.lean`: `recvsAt`/`sendsAt`/`excessAt`/`admissibleMemoryBusM`, `tsSlotVal`/`tsBounded`, `entryKeyed`, the perm sanity theorems, and the header docstring carrying the balance + atomicity justification. The positional predicate remains temporarily (consumed by nothing audited).
- `ApcOptimizer/OpenVmSemantics.lean`: `memTsFieldOf` (memory → slot 6, exec bridge → slot 1), `memEntryKeyOf` (exec bridge → slot 0, the caller's entry pc), and the four-conjunct `admissible`. Trust claims to check: the payload slot layout, OpenVM's global `timestamp < 2^29` argument, and the entry pc.
- `ApcOptimizer/Sp1Semantics.lean`: the two-conjunct SP1 `admissible`; SP1 declares **no** timestamp slot (its clock is two payload fields, shard/clk) and no entry key, documented in place.
- `ApcOptimizer/Optimizer.lean`: `openVmOptimizer` and its correctness theorem gain the `entryPc` parameter — quantified over, so the theorem ties the optimizer to the semantics judged against *the same* value. The value is bound outside the audited files (`Ffi.lean`, `Main.lean`), like `busMap`.

Everything else (~2,600 new lines under `Implementation/`: the forced-cascade combinatorics, the multiset consumption theorems, the chain lemma, the value-level no-wrap lemmas, the rebuilt `busUnify`, the new `execChain`) carries its own machine-checked proofs and needs no audit per this repository's architecture.

## How the passes work under the new rely

- `busPairCancel`: unchanged in power. Its `admissible_dropPair` obligation *simplified* — the shield and region conditions are gone; dropping an equal-payload matched pair leaves every per-address excess multiset unchanged (`excessAt_dropPair`), which also carries the entry designation.
- `busUnify`: rebuilt on a timestamp-group engine. Per bus and address group it certifies fiber completeness (every interaction classified by the address-disequality certificates; any undecided interaction conservatively aborts the group), a shared syntactic timestamp base, and the inlined LessThan gadget (certified through the range-checker facts). Sends are ordered by the **differences** of their timestamp constants, not their absolute offsets (`val_lt_of_step`): the surviving base after substitution is whichever instruction's clock Gauss kept, so the offsets are signed.
- `execChain` (new): certifies the execution bridge's chain — constant key slots that step through the block, one positive constant timestamp step per access — and emits the CPU-state copy equalities. `entryKeyed_chain_copies` consumes ENTRY_KEY, plus the fact that the partner relation strictly decreases receive timestamps and therefore cannot cycle, which is what excludes the exit send (whose jump target is a `jalr` register value or a branch back to the entry) from stealing an interior receive. `gauss` then substitutes the chain, bringing every instruction's memory timestamps onto one base — the precondition cross-instruction memory forwarding needs.
- `busSweep` (new): the pre-migration consecutive-match sweep, restored order-free. It *proposes* the bus's accesses as index pairs — bucket the interactions by canonical address key, pair the k-th receive with the k-th send of each bucket, order the accesses by send timestamp — and then verifies the proposal: the flattened indices sort to `List.range n` (so the accesses cover every interaction exactly once), each pair carries `denseBUConstsEq` on the address slots plus the solved LessThan gadget, and the sends' timestamps share one linear base whose constants step by `[1, B)` along the proposed order. The permutation certificate transports `admissibleMemoryBusM` to the *rearranged* list (`admissibleMemoryBusM_perm`), where `admissibleMemoryBus_of_pairUp` (`Implementation/MemoryBusOrdered.lean`) yields the positional discipline as a theorem — `pairUp` of a `flatMap (fun q => [q.1, q.2])` is the identity, so the chunking hypothesis is free by construction. On that list the old sweep, its window bookkeeping, its mid-scan and its `denseCheckPair` verifier run unchanged; the one line of the old proof that consumed the positional rely now points at the bridge instead.

## Results

| suite | variables (main → PR) | bus interactions (main → PR) |
|---|---|---|
| openvm-eth | 4.553× → **4.553×** | 3.558× → **3.557×** |
| wasm-eth | 7.254× → **7.254×** | 6.211× → **6.205×** |
| sha256 | 14.191× → **14.191×** | 8.735× → **8.734×** |
| keccak | 13.618× → **13.618×** | 7.587× → **7.587×** |

Full parity on every suite. keccak's output is identical to `main`'s (2021 variables, 186 constraints, 1748 bus interactions). Local `Benchmarks/benchmark.py`; CI numbers on this push should match.

### On the aliasing story in the previous revision of this description

The earlier text here claimed the remaining `sha256`/`keccak` gap *was* the aliasing wall — that the order-free consumption needs a group's exact fibers, and the 54 symbolic-pointer groups therefore could not be certified. That was an inference from the surviving-interaction census, and measurement does not support it. The census is right (540 memory interactions survive keccak, 482 of them in those groups); the causal story was wrong.

Instrumenting both engines at the cycle where they fire shows the timestamp-group engine emits **exactly** what the consecutive-match sweep emits — 9272 raw / 7920 new equalities on keccak, identical fixpoint. A first version of `busSweep` built on list adjacency (send at position *i*, its consumer at *i+1*) reproduced that number and changed nothing at all. At that point in the pipeline the group engine *can* certify the symbolic-pointer groups; aliasing is not what stops it.

The actual gap was in the *late* cleanup cycles. Per-cycle against `main` on keccak:

| cycle | main (vars / bus) | this branch, before (vars / bus) |
|---|---|---|
| 2 | 5428 / 3610 | 5443 / 3622 |
| 3 | 5013 / 3216 | 5081 / 3261 |
| 5 | 2795 / 2350 | 3045 / 2772 |
| 7 | **2021 / 2056** | **2867 / 2586** |

From cycle 3 the memory bus froze here and kept shrinking on `main`. The reason is `busPairCancel`: once it has removed the matched pairs, a surviving receive and its partner send are no longer *neighbours* in the list (125 of 299 pairs adjacent on keccak), so any adjacency-based chunking fails and both engines go dead — while the positional rely, being assumed rather than certified, keeps `main`'s sweep working on the thinned list.

Pairing by **matching** instead of by adjacency is what closes it: bucketing by address key and pairing the k-th receive with the k-th send certifies 299/299 pairs where adjacency certified 125. A receive whose own access's send has been cancelled is still certifiable against a *later* same-address send, because its timestamp slot is still the lt-aux decomposition `(base + c₁) − 1 − Σ limbs` of the shared base, so `send.ts − recv.ts` normalizes to `(c_k − c₁ + 1) + Σ limbs` — a `c₀ ≥ 1` constant over the same range-checked limbs, exactly the shape `denseBUGadgetOk` recognizes. The gadget is a certificate about any later same-base send, not only about its own access's.

This is also why the pass never has to decide aliasing. `interleaveAccesses_admissibleMemoryBus_of_M` does not require the pairs to be *real* accesses; it needs only that the pairs' components are the bus's send/receive fibers, that each pair shares an evaluated address, that each receive's timestamp is below its send's, and that the send timestamps increase along the list. Any perfect matching meeting those works, and each evaluated-address group falls out as a *subsequence* of the global order — so no interaction is ever certified disjoint from another. The 54 symbolic-pointer groups simply stop being the unit of reasoning.

Known follow-up on this branch: runtime. `sha256` costs +15% (296 s vs 257 s), but keccak is 8.2 s against `main`'s 1.2 s. `denseBUVarBound` rescans every bus interaction per limb variable, once per candidate pair; memoizing it into a `VarId → Nat` map built once per invocation is the fix, and it is worth checking whether the timestamp-group engine is now subsumed by `busSweep` altogether.

Also unchanged by this PR: SP1 declares no timestamp slot, so `busUnify` does not fire there at all and the SP1 suites remain below `main`. Recovering them needs a two-field-clock rely, tracked separately.

`Scripts/check-proof-integrity.sh` and the build are green, with no `sorry`/`axiom`/`native_decide` and the correctness theorems resting only on Lean's three standard axioms.


1arie1:
Status update after the CI benchmark: the effectiveness regression is **universal across suites** (e.g. keccak 13.6× → 2.0× on variables, sha256 runtime +331%), not the keccak-only effect the original description claimed — the PR body has been corrected accordingly. Root cause: the timestamp-based `busUnify` only certifies same-instruction address groups; cross-instruction forwarding needs the exec-bridge chain equalities, which are recoverable under the same order-free rely via constant-pc-pinned matching (plan in the updated body). Keeping this as a draft until that lands; the audited-surface redesign and all proofs remain green and unaffected.

ApcOptimizer/Sp1Semantics.lean
```diff
@@ -176,14 +176,31 @@ def sp1BusSemantics (p : ℕ) (busMap : BusMap := defaultBusMap) :
...
+  -- address, the excess of the `getPrevious` payload multiset over the `setNew` one.
+  -- No TS_BOUND (`tsBounded`) conjunct is declared for SP1: its clock is *two* payload fields
+  -- (`clk… (2 fields)` on both memory and the execution bridge — see the payload docstrings
+  -- above), not a single bounded timestamp slot, so timestamp-order reasoning stays unavailable
+  -- for SP1 until a single-slot encoding of its clock is established.
   admissible msgs :=
     (∀ (busId : Nat) (shape : MemoryBusShape), memShapeOf busMap busId = some shape →
-      admissibleMemoryBus shape (msgs.filter (fun m => m.busId = busId)))
+      admissibleMemoryBusM shape
+        (↑(msgs.filter (fun m => m.busId = busId)) : Multiset (BusInteraction (ZMod p))))
```
georgwiese, line 189: The `tsBounded` property is missing here. And I think it's actually kind of complicated in the SP1 case:
- The time stamp is two limbs, not one
- The have these concepts of "memory bump" and "state bump". I'd have to look up the exact details, but basically a circuit only increments the lower limb (keeping the higher limb unchanged), which might cause it to go out of bounds. The prover is forced to correct this by nondeterministically invoking a chip that receives the out-of-bounds time stamp, does the carry (changing the higher limb) and sends it back to the bus.

ApcOptimizer/MemoryBus.lean
```diff
@@ -68,3 +85,46 @@ def admissibleMemoryBus (shape : MemoryBusShape) (L : List (BusInteraction (ZMod
...
+    exceeds the sends' payload multiset by at most one element — the entry receive. -/
+def admissibleMemoryBusM (shape : MemoryBusShape)
+    (M : Multiset (BusInteraction (ZMod p))) : Prop :=
+  ∀ addr : List (Option (ZMod p)),
+    Multiset.card
+      ((recvsAt shape addr M).map BusInteraction.payload
+        - (sendsAt shape addr M).map BusInteraction.payload) ≤ 1
+
+/-- The discipline is invariant under reordering the interaction list. -/
+theorem admissibleMemoryBusM_perm (shape : MemoryBusShape)
```
georgwiese, line 118: I generally kept theorems out of the spec, because they just state true facts, not assumptions that have to be audited. Was it a deliberate choice to have the theorem in this file, as opposed to the implementation section?

georgwiese (commented):
Cool! I'm starting to get the idea. My understanding is that the assumption that memory interactions are ordered by time is replaced by two assumptions:
1. Time stamps are bounded (e.g. $< 2^{29}$ in OpenVM).
2. All receives can be matched to a send, except one.

I think from these assumptions, there could be an algorithm that restores the order:
- Sends can be ordered if all sent time stamps are relative to the same "base" time stamp (which should be the case in practice). This is where assumption (1) comes in as well, to rule out wrapping.
- Receives can be matched to sends by reasoning about the `received_ts < base_ts + <constant>` assertions that the input circuit should always have. This is where assumption (2) comes in, which forces all but one to be matched. Possible, but sounds complicated to reason about, because this less-than assertion is not very direct.

I think it should be possible to have the exact same optimizer (same effectiveness and runtime), but just put in a step that orders memory bus interactions (possibly increasing runtime slightly).

But taking a step back, I want to understand the premise of this PR better:
> The old rely couples soundness of memory-pass completeness to an unverifiable frontend property ("interactions are listed in chronological order"). The new rely is stated against facts an auditor can check independently of emission order, and is strictly closer to definitional properties of the bus argument.

Why is it an unverifiable property? I think to verify it, one has to:
1. Verify that each instruction circuit lists memory bus interactions in time order. For OpenVM, this involves going through each of the 13 chips [here](https://github.com/powdr-labs/powdr/blob/5aa140d35a9cc8467be286fb44969cfbab3533bb/openvm-riscv/tests/openvm_constraints.txt).
2. Verify that we concatenate instruction circuits in the order in which these instructions are executed, when building the input to the optimizer. That involves auditing [this function](https://github.com/powdr-labs/powdr/blob/5aa140d35a9cc8467be286fb44969cfbab3533bb/autoprecompiles/src/symbolic_machine_generator.rs#L132-L146).

That doesn't sound much more complicated than validating assumption (2) above?

ApcOptimizer/MemoryBus.lean
1arie1: This is not a real theorem in the sense that it is not used anywhere. It is a test for the spec. It shows that an admissibility is not affected by permutation. It should definitely be moved to some "test" module.

ApcOptimizer/Sp1Semantics.lean
1arie1: I don't know much about SP1 and have ignored it for now. Is there a white paper / documentation to learn about how it encodes the time stamps?

1arie1:
@georgwiese one correction to your first comment. After the sends are ordered, a recv does not need to be matched with a corresponding send. This might not be possible due to unknown aliasing. However, the recvs can be ordered by their upper bound constraints. Then, a recv `R` that is directly bounded by a send `S` is always receiving the latest previous value.

1arie1:
@georgwiese I've updated this PR with better effectiveness. This brought it one extra assumption -- the pc that is sent to the block being optimized is the pc of the first instruction of the block. This seems like a reasonable (and true) assumption. Without the assumption, there are issues for blocks that are self-loops. In which case, all instructions in a block can be executed even when the block is entered from the middle. This prevents memory bus optimization across instructions (and negatively impacts effectiveness)

1arie1:
@georgwiese for your other concern, maybe it is simpler to discuss this in our meetings. "Unverifiable" is always subject to what we assume is in scope. The current formulation, for example, relies mostly on the OpenVM white paper. The order of bus interactions is forced by the constraints. In the alternative formulation, the assumption relies on OpenVM implementation producing bus interaction in a specific order, and, everything else in the pipeline, preserving that original order.

1arie1:
Pushed `b2d3e24`: the consecutive-match sweep is back, order-free, and all four suites are now at parity with `main` (keccak `9.599x -> 13.618x`, sha256 `12.172x -> 14.191x`, openvm-eth `4.553x`, wasm-eth `7.254x`; keccak’s output is identical to `main`’s).

Two corrections to what this description said before, both from measurement rather than reasoning:

1. The remaining gap was **not** the aliasing wall on the symbolic-pointer groups. At the cycle where they fire, the timestamp-group engine and the consecutive-match sweep emit exactly the same equalities (9272 raw / 7920 new on keccak); a first version of `busSweep` built on list adjacency changed nothing at all.
2. The gap was in the **late** cleanup cycles. Once `busPairCancel` has removed the matched pairs, a surviving receive and its partner send are no longer neighbours (125 of 299 pairs adjacent on keccak), so adjacency-based chunking dies — while the assumed positional rely keeps `main`’s sweep working on the thinned list. Pairing by *matching* rather than adjacency certifies 299/299 and closes it.

`admissibleMemoryBus_of_pairUp` is now applied to a **reordered** list: the pass proposes index pairs, verifies that they sort to `List.range n`, and transports the rely across the permutation with `admissibleMemoryBusM_perm`. So this also answers the ordering question from the review thread concretely — the "step that orders memory bus interactions" is in the pass, it is verified rather than trusted, and the receive placement is exactly the solved LessThan gadget (`denseBUGadgetOk_sound`, `val_lt_of_lessThan_gadget`) that was already in the tree.

Still open: runtime (keccak 8.2 s vs `main`’s 1.2 s; sha256 only +15%), and SP1, which declares no timestamp slot.

1arie1:
@georgwiese would it be possible to get new bench run on the current state of the PR. It should be closer to `main` in effectiveness now

georgwiese:
I'll close and reopen to see if that triggers CI...

ApcOptimizer/MemoryBus.lean
```diff
@@ -68,3 +90,69 @@ def admissibleMemoryBus (shape : MemoryBusShape) (L : List (BusInteraction (ZMod
...
+def sendsAt (shape : MemoryBusShape) (addr : List (Option (ZMod p)))
+    (M : Multiset (BusInteraction (ZMod p))) : Multiset (BusInteraction (ZMod p)) :=
+  M.filter (fun m => m.multiplicity = shape.setNewMult ∧ shape.address m = addr)
+
+/-- The payloads the receives at `addr` hold in excess of the sends: what enters the block from
+    outside there. `admissibleMemoryBusM` bounds its cardinality by one. -/
+def excessAt (shape : MemoryBusShape) (addr : List (Option (ZMod p)))
+    (M : Multiset (BusInteraction (ZMod p))) : Multiset (List (ZMod p)) :=
+  (recvsAt shape addr M).map BusInteraction.payload
+    - (sendsAt shape addr M).map BusInteraction.payload
```
leonardoalt, line 109: what if recvs payloads = {A, B, C}, and send payloads = {B, C, D}, are the extras only relative to recvs, i.e. D is ignored?

1arie1: Yes. The assumption is only for incoming data. So for every payload, there is at most one incoming value. 

There is no assumption on outgoing values. However, by soundness, the number of outgoing unmatched sends in the optimized circuit must be equal the number of outgoing unmatched sends in the original circuit.

ApcOptimizer/Sp1Semantics.lean
georgwiese: Not that I'm aware of, unfortunately.

This is a summary of how they handle time stamps, extrapolating from the [`ADDI` chip](https://github.com/powdr-labs/powdr/blob/ca07f72427a2365db797e052c175161f91a8b667/sp1-benchmarks/tests/sp1_constraints.txt#L928-L1004):
- From the execution bridge, they receive two limbs timestamp limbs, into expressions `(state__clk_high, state__clk_0_16 + state__clk_16_24 * 65536)` ([source](https://github.com/powdr-labs/powdr/blob/ca07f72427a2365db797e052c175161f91a8b667/sp1-benchmarks/tests/sp1_constraints.txt#L984))
- `state__clk_high` is not changed at all. Whatever value is received from the execution chip [is also sent back to it](https://github.com/powdr-labs/powdr/blob/ca07f72427a2365db797e052c175161f91a8b667/sp1-benchmarks/tests/sp1_constraints.txt#L985). Also, it is used in both the [memory bus receives and sends](https://github.com/powdr-labs/powdr/blob/ca07f72427a2365db797e052c175161f91a8b667/sp1-benchmarks/tests/sp1_constraints.txt#L963-L966).
- `state__clk_0_16` and `state__clk_16_24` are range-checked via the `BYTE` bus. Basically, the lower limb must be a 24-Bit value which equals 1 mod 8 (instructions typically increment the time stamp by 8).
- In the [execution bus send](https://github.com/powdr-labs/powdr/blob/ca07f72427a2365db797e052c175161f91a8b667/sp1-benchmarks/tests/sp1_constraints.txt#L985), the lower limb is sent as `state__clk_0_16 + state__clk_16_24 * 65536 + 8`.
- So, reading this chip in isolation, there are two problems:
  - 1. What if there needs to be a carry? The `ADDI` chip just increments the lower limb by 8, which means that the new value might be out of range. As a consequence, the range checks on the next instruction would fail.
  - 2. Assuming the carry does happen somehow, how can `state__clk_high` be hard-coded in the memory bus interactions that "get" the previous time stamp? (Annoyingly, SP1 inverts the order here: previous values / time stamps are *sent* and new ones *received*.) What if the memory cell was last accessed before the last carry event?
- The two issues are solved using two chips that can be invoked by the prover at any point in time.
  - 1. [`StateBumpChip`](https://github.com/succinctlabs/sp1/blob/powdr-labs/apc-air-and-executor/crates/core/machine/src/adapter/bump.rs#L185-L247): Receives the current state from the execution bus (which might have an out-of-range lower limb), does the carry and sends the updated state.
  - 2. [`MemoryBumpChip`](https://github.com/succinctlabs/sp1/blob/powdr-labs/apc-air-and-executor/crates/core/machine/src/memory/bump.rs#L179-L207): Does a read access. The address and time stamp can be chosen by the prover. With this, the prover can bump the time stamp (only increases are allowed), while keeping the values the same.

BTW, something similar happens for increments to the PC, which is also spread across several limbs. The `ADDI` instruction [only increments the lowest limb by 4](https://github.com/powdr-labs/powdr/blob/ca07f72427a2365db797e052c175161f91a8b667/sp1-benchmarks/tests/sp1_constraints.txt#L985). Any carry also happens via the `StateBumpChip`.

georgwiese: In our SP1 integration, we handle all this by not using the APC at all if any of these carry events happen (which are rare). This works because the prover is generally not forced to use the APCs. Put another way, we're not 100% complete, namely the APCs are not satisfiable in these carry events.

1arie1: `state_clk_high` is hard-coded to force register access to stay within an epoch. If an epoch switch happens during a register access, then `MemoryBumpChip` transports the required register value from prior epoch to the current one.

1arie1:
SP1 support landed on this branch (`ef4689e`…`dadabc28`) and is up for CI as the stacked #303 — TS_BOUND on `clk_low` under the single-epoch argument, ENTRY_KEY by the entry pc's low limb, and the gadget certificates extended to SP1's byte-bus range-check shapes. Local numbers: rsp 1.838x → 3.868x (main 3.922x), sp1:keccak 1.704x → 5.042x (now ahead of powdr's 4.809x), OpenVM keccak unchanged.

ApcOptimizer/MemoryBus.lean
```diff
@@ -68,3 +90,69 @@ def admissibleMemoryBus (shape : MemoryBusShape) (L : List (BusInteraction (ZMod
...
+
+/-- ENTRY_KEY: every payload entering the block from outside carries `key` in slot `slot`.
+
+    For a chain bus (an execution bridge), window atomicity already says *one* record enters; this
+    says it is the block's entry record, identified by its key — the block is entered at its entry
+    pc, which the optimizer is told. The designation cannot be recovered from the message multiset:
+    a *rotation* of the block's records (entered at an interior instruction, wrapping through the
+    exit) yields the same multiset with a different entry, so it takes an assumption to exclude. -/
+def entryKeyed (shape : MemoryBusShape) (slot : Nat) (key : ZMod p)
+    (M : Multiset (BusInteraction (ZMod p))) : Prop :=
```
georgwiese, line 132: Out of curiosity, I think the use of `Multiset` is a bit strange here, because `BusInteraction` itself has a `multiplicity` field? I think we'd want to model a multiset of `(busId, payload)` elements, but now it's a multiset of `(busId, payload, multiplicity)`.

In the main spec, we modeled the `BusState` as `BusMessage p → ZMod p` (where `BusMessage` is `Nat × List (ZMod p)`), I'll probably try to refactor it in a way that's more similar to this. I do think that the current formalization still captures what we want though.

Pls let me know if I'm overlooking something :)

1arie1: What are you missing is that this is a syntactic condition on the presentation of the circuit, rather than semantic on the execution. See how it is used here:

https://github.com/1arie1/apc-optimizer/blob/dadabc2854736d7e6c75ac4d2cbb4ff15ff092f1/ApcOptimizer/OpenVmSemantics.lean#L217

Syntactically, a circuit has a list of `BusInteraction`. However, we want stress that the order of this list is not important. Hence, we convert it to a `Multiset`. So think of this `Multiset` as `a list without an order`. 

Then, the syntactic admissibility check is that all sends and receives in this `Multiset` balance out, except for possibly one extra receive and some extra sends. 

So the condition is on the syntax -- the input circuit has a bunch of bus interactions with given multiplicity, but without any specific order. Hence, it is a `Multiset` rather than the balancing condition on the multiplicities.

hope that clears it up

georgwiese:
@1arie1 I agree with the spirit of the spec change and would like to do some small iterations on the spec and implementation. Could you change the target branch to `order-free-admissibility` instead of `main`? I also invited you to be a collaborator to the main repo, so you'd still be able to push to the new branch.

georgwiese:
I played around with the PR a little bit and added some commits on top, see #304. These are some of the changes:
1. Refactored the spec a little bit. It might be a matter of taste, but I find it more readable. It reuses the notion of `BusState` from the main spec, and explicitly states that for each address either (1) the memory bus is fully balanced, or (2) there is exactly one entry and one output record. This is a stronger statement than in this PR (which doesn't fix the number of output records), but I think it follows the same spirit in that it can be verified just from the white paper, without assumptions on the order of constraints or anything like that.
2. Pushed Claude to reduce the net additions from 4.5k -> 1.5k LoC.
3. Pushed Claude to improve effectiveness and runtime.

The main thing that is still a bit ad-hoc is this entry PC thing. I think it's unfortunate for a few reasons (adds an extra input, hard-codes more specifics about the execution bus, etc), but the biggest one is that I believe it would break with branch prediction: The optimizer currently assumes that all received PCs are unique. In my understanding, if it wasn't, the execution bus interactions couldn't be matched using just the initial PC as input, which would lead to a massive effectiveness drop.

Currently, PCs are guaranteed to be unique, because we only generate APCs for basic blocks, so they would always increment by 4. But we do have experiments with branch prediction, where for example we might want to compute an APC that unrolls two iterations of a loop. My understanding is that in that case, we'd actually have to pass the full order of execution bus interactions again, same as we do (implicitly) on `main`!

I discussed this with @Schaeff today and he made the point that the right thing to do would be to pull in [this code](https://github.com/powdr-labs/powdr/blob/f94a24f19249af67efbea92ff9c3db6e3e50e7fd/autoprecompiles/src/symbolic_machine_generator.rs#L134-L205) (the Rust code building the input circuit from the basic block (or super block, in the case of branch prediction) and the instruction circuits) into the Lean optimizer. I think this would be a good idea anyway, because now it is extra code that needs to be audited. If this code was part of the Lean APC builder, we could *prove* that instruction circuits are concatenated in the order of execution. So basically:
- Within an instruction circuit, we wouldn't assume any order. The assumptions on the memory bus would be like in this branch and can be verified by checking against the white paper.
- Inter-instruction we would assume an order, and this assumption can be proven. With this, the execution bus can be processed like it is on `main`.

PR #290, opened 2026-08-05
perf(guard): lockstep degree check with withPtrEq identity shortcuts (0.94-0.96x end-to-end)

pacheco:
## What

`guardDegree` re-ran `withinDegreeB` — a full degree walk over every item of the system — on every pass output, including the ~62% of invocations that return their input untouched (the R5/R15 measurements). The check now runs **in lockstep against the pass input**:

- `withPtrEq` identity shortcuts at the whole-system and per-item level skip the walk for anything that *is* the input object (a no-op pass costs one pointer compare; a rebuilt-spine pass with preserved items costs one pointer compare per item).
- Each paired position tests `outOk || !inOk`. That is `true` on identical items by Bool excluded middle — which is exactly `withPtrEq`'s proof obligation, so the change is **axiom-free**: no `unsafe`, no `@[implemented_by]`, and the integrity script still reports `[propext, Classical.choice, Quot.sound]` on the correctness theorems.
- On the within-bound inputs the guarded pipeline maintains, `outOk || !inOk` decides exactly `outOk` (`denseWithinDegreeLK_sound`, consumed by `guardDegree_respectsDeg`). Unpaired output items (appends, length changes) get the plain check, so the guard never rejects a within-bound output.

`denseWithinDegreeLK` is `@[irreducible]`: without it, `optimizerWithBusFacts_correct`'s `change` tactics whnf through the folded 40-pass pipeline into the guard check and blow past 16x the heartbeat budget; with it, no options change anywhere.

## Measured (interleaved A/B vs main, 3 reps + one sha256 shot, serial)

| case | main | this PR | ratio |
|---|---:|---:|---|
| sha256 apc_001 | 20.79 s | 20.01 s | **0.962x** |
| wasm-eth apc_012 | ~2.43 s | ~2.29 s | **0.940x** |
| keccak apc_001 | ~2.05 s | ~1.97 s | **0.963x** |
| wasm-eth apc_036 | ~2.13 s | ~2.02 s | ~0.95x |

Per-pass, the rebuilt cheap sweeps that were mostly guard collapse: degenRange 22→7 ms, hintCollapse 15→9, bytePack 18→10 on the representatives; sha256 flagFold 1094→1028, domainFold 1191→1124.

**Effectiveness:** expected unchanged — the check is weaker-or-equal per item and never rejects a within-bound output, so decisions are identical on invariant-respecting runs. Output verified identical on wasm-eth apc_012/apc_036, OpenVM keccak, SP1 keccak, and sha256's final sizes; the CI matrix is the full verdict.

## Bookkeeping

`agent-docs/log.md` entry 181 records this change plus three dead ends measured in the same session (shrink-first pre-cycle: negative, gauss basis-sensitivity; guard de-boxing without lockstep: flat — the guard's cost is the walk's memory latency, not the loop machinery; reencode seed capacity slack: flat). `ideas.md` R5 and the dead-ends list updated accordingly. The remaining half of R5 (the `unchanged` field, fixpoint `sizeKey` skip) is deliberately out of scope: sub-bar alone now, and the substrate for the R15 cross-cycle skip memo to be sized with it.

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/Pass.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/Pass.lean
@@ -86,13 +86,151 @@ def DenseRespectsDeg (b : DegreeBound) (f : DenseVerifiedPassW p) : Prop :=
     (reg.decodeCS d).withinDegree b →
     ((f reg d hcov bs facts).reg'.decodeCS (f reg d hcov bs facts).out).withinDegree b
 
+/-! ### The lockstep degree check
+
+The guard's degree walk runs against the pass *input* in lockstep: an output item that *is* the
+aligned input item (the same object, `withPtrEq`) is skipped without walking either tree. What each
+paired position checks is `outOk || !inOk` — `true` whenever the two items are equal
+(`bool_or_not_self`, which is exactly `withPtrEq`'s contract) and equal to the plain check whenever
+the input side is within the bound, which the guarded pipeline maintains
+(`denseWithinDegreeLK_sound`). Unpaired output items (appends, length changes) get the plain
+check. -/
+
+private theorem bool_or_not_self (b : Bool) : (b || !b) = true := by cases b <;> rfl
+
+/-- One lockstep item check: `outOk || !inOk`, with the identity shortcut. -/
+def denseDegItemLK (bnd : Nat) (cIn cOut : DenseExpr p) : Bool :=
+  withPtrEq cOut cIn
+    (fun _ => decide (cOut.degree ≤ bnd) || !decide (cIn.degree ≤ bnd))
+    (fun h => by subst h; exact bool_or_not_self _)
+
+private theorem denseDegItemLK_eq (bnd : Nat) (cIn cOut : DenseExpr p) :
+    denseDegItemLK bnd cIn cOut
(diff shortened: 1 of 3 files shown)


pacheco:
**Updated: rebased onto current main and upgraded with the two walk refinements from #281, which this PR now supersedes.**

A three-way interleaved A/B (main vs #281 vs this branch, both rebased onto the same main, outputs bit-identical everywhere) showed #281's edge came from its walk, not its certificate: the remaining-list `withPtrEq` (one pointer compare retires a shared tail) and monomorphic per-item walkers (no closure allocation per interaction). Both are now ported here, certificate-free: `denseDegItemsFast`/`denseDegBisFast` carry their own spec (`{ r : Bool // r = denseDegItemsLK … }`) so the list-level `withPtrEq` obligation is discharged by the walk's `_self` lemma mid-recursion — no threaded proof, no framework type change, still one file.

Measured (medians, vs current main): **sha256 0.940×**, apc_036 0.933×, keccak 0.931×, apc_012 0.962×. #281 respectively: 0.948–0.956× / 0.959× / 0.954× / 0.922× — a statistical tie on aggregate, this branch ahead on three of four cases including sha256, #281 ahead on apc_012 only (noted in its close-out). Proof-integrity clean.

PR #291, opened 2026-08-05
perf: three no-op-prologue fixes (domainFold, carryBranch, reencode)

pacheco:
Three contained fixes to work passes do before discovering they have nothing to do. From a recognizer-level analysis of the no-op wall (~30% of cleanup pass time is passes finding nothing; cross-pass gating and cross-cycle memoization of that wall were both prototyped and measured dead — the surviving lever is making each pass's own discovery cheaper). Log entry 181 has the full story including two measured dead ends recorded in ideas.md.

- **domainFold** (`dfPlan`): bail before the ≤256-point box enumeration when `dfCovScan` already found no fold position and no touched interaction — the collect results were empty regardless of the survivors. **0.94–0.97× on the pass.** Proof: two extra `split` arms.
- **carryBranch** (`denseResolveExpr`): the unresolved-product fallthrough rebuilt `.mul f g` — ~157k reallocated roots per sha256 invocation, so no unchanged constraint was ever pointer-identical to its input. `e@(.mul f g) => … else e` returns the original node; definitionally transparent, proofs untouched. **0.97–0.99× now**, and it re-enables per-item pointer shortcuts in the lockstep degree guard (#290) when that lands.
- **reencode** (`denseRncBuild`): the covered-set gather (`denseCoveredIdxPos`: bucket union → HashSet → mergeSort → per-position `denseCoveredBy` walk) ran before the domain lookup and box gate that reject a group for free. Reordered — with the gather reading the post-doms state: a pending use of the pre-state un-shares the memo arrays inside `denseRncDoms` and turns its in-place updates into full copies (the first version measured 2× on the pass before the aliasing fix). **0.97–1.00× on the pass.** Proof: `denseRncBuild_spec` rewrites through `DenseRncCore`'s anchor/cs preservation.

**Measured** (interleaved A/B, 3 reps + sha256 shots): wasm apc_012 **0.990×**, apc_036 **0.993×**, keccak **0.995×**, sha256 neutral. Output bit-identical on all four spot cases; `Scripts/check-proof-integrity.sh` clean. Effectiveness is unchanged by construction (every change only skips work whose results were provably discarded, or reuses a node the pass rebuilt identically).

### DIFF
--- a/ApcOptimizer/Implementation/OptimizerPasses/CarryBranch.lean
+++ b/ApcOptimizer/Implementation/OptimizerPasses/CarryBranch.lean
@@ -115,10 +115,10 @@ def denseNeverZeroB (ops : DenseZModOps p) (B : Std.HashMap VarId Nat) (e : Dens
     becomes `x-5 = 0`. Recurses into the surviving factor. -/
 def denseResolveExpr (ops : DenseZModOps p) (B : Std.HashMap VarId Nat) :
     DenseExpr p → DenseExpr p
-  | .mul f g =>
+  | e@(.mul f g) =>
       if denseNeverZeroB ops B g then denseResolveExpr ops B f
       else if denseNeverZeroB ops B f then denseResolveExpr ops B g
-      else .mul f g
+      else e
   | e => e
 
 theorem denseResolveExpr_vars (ops : DenseZModOps p) (B : Std.HashMap VarId Nat) (e : DenseExpr p) :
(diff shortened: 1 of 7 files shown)


pacheco:
Rebased onto main post-#290 (log entry renumbered 181→182; ideas.md dead-end references updated; the three perf commits applied unchanged). Output remains bit-identical against merged main on all four spot cases; proof-integrity clean.

**Honest re-measurement on top of the merged guard** (interleaved, n=6 + sha256 shots): the end-to-end margin has narrowed to neutral-within-noise — apc_012 1.009×, apc_036 1.008×, keccak 0.993×, sha256 flat. The lockstep guard absorbed much of what these fixes saved in absolute terms (post-guard, the touched passes' totals are already substantially lower, e.g. domainFold 62→50ms on apc_012's main). The three changes remain structurally discard-only (each skips work whose results were provably unused, or reuses a node the pass rebuilt identically), and carryBranch's pointer-stable returns feed the guard's per-item shortcut — but on today's main their measured value is within noise on the small cases. Leaving the merge call to the CI corpus run and the reviewer.

PR #293, opened 2026-08-10
spec: make witness generation take the coverage proof

georgwiese:
`Derivations.witgen` had two comments claiming a branch was unreachable "by `Derivations.cover`". Both are now enforced by types: it takes the coverage proof and `hv : v ∈ outputVars`, reads the method out of `cover`'s existential, and has no fall-through case. It therefore cannot be applied without coverage in hand, so a spec clause cannot silently omit that obligation — `isCompleteReplacementOf` binds it with `∃`.

Since witness generation is now defined exactly on the output circuit's variables, the spec quantifies over the assignments agreeing with it there instead of naming one. No default value is invented anywhere, and the clause is equivalent: `satisfies`/`admissible`/`sideEffects` read nothing outside those variables.

Implementation side: `witgen_powdrId` / `witgen_methodFor` characterize the two branches, so the completeness proof never unfolds `witgen`. Note that a goal mentioning a `witgen` application reduces its implicit `outputVars` under `cases`/`obtain` — i.e. runs the whole optimizer inside `whnf` — so the proof case-splits before rewriting.

Spec-only: no pass, no effectiveness, no runtime behaviour changes. `lake build` is warning-free and `Scripts/check-proof-integrity.sh` passes on the same three standard axioms.

### DIFF
--- a/ApcOptimizer/Spec.lean
+++ b/ApcOptimizer/Spec.lean
@@ -213,11 +213,8 @@ def Derivations.cover (ds : Derivations p)
 -- ANCHOR_END: methodForCover
 
 omit [Fact p.Prime] in
-/-- Restating the `none` branch of `cover`: a covered variable without a powdr
-    ID has a derivation. Only ever used as the `Option.get` obligation in
-    `witgenOn`, so by proof irrelevance the proof below is not part of the
-    specification — it either typechecks or it does not. -/
-theorem Derivations.methodFor_isSome_of_cover {ds : Derivations p}
+/- Support lemma, does not require audit. -/
+theorem Derivations.methodFor_isSome (ds : Derivations p)
     {inputVars outputVars : List Variable} (h : ds.cover inputVars outputVars)
     {v : Variable} (hv : v ∈ outputVars) (hp : v.powdrId? = none) :
     (ds.methodFor v).isSome := by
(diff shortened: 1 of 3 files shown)


ApcOptimizer/Spec.lean
```diff
@@ -211,24 +211,52 @@ def Derivations.cover (ds : Derivations p)
...
-    | none => inputAssignment v
+    assignment. On `outputVars` this is `witgenOn` — what powdr runs to fill the
+    optimized circuit's variables from an input trace. Off `outputVars` the value
+    is arbitrary; the output circuit's constraints and bus interactions cannot
+    read it. -/
+def Derivations.witgen (ds : Derivations p) {inputVars outputVars : List Variable}
+    (h : ds.cover inputVars outputVars) (inputAssignment : Variable → ZMod p)
+    (v : Variable) : ZMod p :=
+  if hv : v ∈ outputVars then ds.witgenOn h inputAssignment v hv
+  else inputAssignment v
```
georgwiese: Could return 0 here?

ApcOptimizer/Spec.lean
```diff
@@ -211,24 +211,52 @@ def Derivations.cover (ds : Derivations p)
...
+def Derivations.witgenOn (ds : Derivations p) {inputVars outputVars : List Variable}
+    (h : ds.cover inputVars outputVars) (inputAssignment : Variable → ZMod p)
+    (v : Variable) (hv : v ∈ outputVars) : ZMod p :=
+  if hp : v.powdrId? = none then
+    ((ds.methodFor v).get (by
+      have hc := h v hv
+      simp only [hp] at hc
+      obtain ⟨cm, hcm, -⟩ := hc
+      simp [hcm])).eval inputAssignment
+  else inputAssignment v
```
georgwiese: You dropped the comment that this case is well defined because by `Derivations.cover`, if `v` appears in the output circuit, it must also exist in the input circuit. I think it still makes sense? I think it's better here than in the general docstring?

ApcOptimizer/Spec.lean
```diff
@@ -211,24 +211,52 @@ def Derivations.cover (ds : Derivations p)
...
+    optimized circuit's variables from an input trace. Off `outputVars` the value
+    is arbitrary; the output circuit's constraints and bus interactions cannot
+    read it. -/
+def Derivations.witgen (ds : Derivations p) {inputVars outputVars : List Variable}
+    (h : ds.cover inputVars outputVars) (inputAssignment : Variable → ZMod p)
+    (v : Variable) : ZMod p :=
+  if hv : v ∈ outputVars then ds.witgenOn h inputAssignment v hv
+  else inputAssignment v
+
+omit [Fact p.Prime] in
```
georgwiese: I think these two theorems could be moved to the implementation section?

ApcOptimizer/Spec.lean
```diff
@@ -294,16 +322,18 @@ def Circuit.isCompleteReplacementOf
   (∀ derivation ∈ ds, derivation.1 ∈ optimizedCircuit.vars) ∧
 
   -- The optimized circuit variables can be derived from the original circuit
-  -- variables, and the return derivations.
-  ds.cover originalCircuit.vars optimizedCircuit.vars ∧
+  -- variables, and the return derivations. This is an `∃` (not a `∀`, which
+  -- would make everything below vacuous): coverage is an obligation on the
+  -- optimizer, and witness generation cannot even be stated without it.
+  ∃ hcover : ds.cover originalCircuit.vars optimizedCircuit.vars,
```
georgwiese, line 304: I don't find the comment here about `∃` necessary. It's quite obvious, and this is just how you give a name to the proof?

leonardoalt (approved):

PR #295, opened 2026-08-10
Encode assumption in circuit type

georgwiese:


ApcOptimizer/Implementation/Optimizer.lean
```diff
@@ -409,10 +454,22 @@ theorem optimizerWithBusFacts_correct {bs : BusSemantics p} (b : DegreeBound) (f
...
 theorem optimizerWithBusFacts_respectsDegree {bs : BusSemantics p} (b : DegreeBound)
-    (facts : BusFacts p bs) (cs : Circuit p)
-    (h : cs.withinDegree b) :
-    (optimizerWithBusFacts b facts cs).1.withinDegree b :=
-  pipeline_respectsDeg b cs bs facts h
+    (facts : BusFacts p bs) (inputCircuit : InputCircuit p)
+    (h : inputCircuit.withinDegree b) :
+    (optimizerWithBusFacts b facts inputCircuit).1.withinDegree b :=
+  pipeline_respectsDeg b inputCircuit.toCircuit bs facts
+    ((CircuitG.withinDegree_mapVar _ inputCircuit b).mpr h)
```
Schaeff: just checking this diff is fine

ApcOptimizer/Implementation/Variable.lean
```diff
@@ -14,14 +14,18 @@ instance : Ord Variable := ⟨fun a b =>
...
+    an id-less name means the input is not what the optimizer's input type describes. -/
+def InputVariable.ofPowdrName (raw : String) : Except String InputVariable :=
   match raw.splitOn "@" with
   | [base, id] =>
       match id.toNat? with
-      | some n => { name := base, powdrId? := some n }
-      | none => { name := raw }
-  | _ => { name := raw }
+      | some n => .ok { name := base, id := n }
+      | none => .error s!"variable without a numeric powdr id: {raw}"
```
georgwiese, line 27: Maybe we can open a PR on powdr that changes the `apc-optimizer` commit to this branch, to test whether this ever happens in the tests?

ApcOptimizer/Spec.lean
```diff
@@ -18,22 +18,44 @@ structure Variable where
...
+    carries a powdr ID: the input circuit has no derived variables. -/
+structure InputVariable where
+  /-- The display name of the variable. -/
+  name : String
+  /-- The powdr variable ID. -/
+  id : Nat
+  deriving DecidableEq, Repr
+
+/-- An arithmetic expression over variables of type `V` and field constants. -/
+inductive ExpressionG (V : Type) (p : ℕ) where
```
georgwiese: I know I chose the name, but I don't like the `G` suffix ^^ I think maybe we can rename:
- `ExpressionG` -> `Expression`
- `Variable` -> `OutputVariable`
- `Expression` -> `OutputExpression`
- `CircuitG` -> `Circuit`
- `Circuit` -> `OutputCircuit`
- `toCircuit` -> `toOutputCircuit`

ApcOptimizer/Spec.lean
```diff
@@ -18,22 +18,44 @@ structure Variable where
   powdrId? : Option Nat := none
   deriving DecidableEq, Repr
 
-/-- An arithmetic expression over structured variables and field constants. -/
-inductive Expression (p : ℕ) where
+/-- A variable of a circuit as exported by powdr. Unlike a `Variable`, it always
+    carries a powdr ID: the input circuit has no derived variables. -/
```
georgwiese:
```suggestion
/-- A variable of a circuit as exported by powdr. -/
```
Sounds like AI explaining the diff of this PR in the comment...

Also, I think we should adjust the `Variable` docstring above, and maybe list `InputVariable` first.

While we're at it, we could also turn `Variable` into an inductive type with two variants, no? This could also be a future PR.

ApcOptimizer/Spec.lean
```diff
@@ -288,14 +337,14 @@ def Circuit.isSoundReplacementOf (optimizedCircuit originalCircuit : Circuit p)
 -- ANCHOR_END: isSoundReplacementOf
 
 -- ANCHOR: isCompleteReplacementOf
-/-- Whether an optimized circuit is a complete replacement for an original circuit. -/
+/-- Whether an optimized circuit is a complete replacement for an original
+    circuit. `Optimizer.isCorrect` demands this only of circuits exported by
+    powdr, whose variables all carry a powdr ID — the ones witness generation
+    reuses from the input assignment. -/
```
georgwiese:
```suggestion
    circuit. -/
```

ApcOptimizer/Spec.lean
```diff
@@ -338,23 +387,23 @@ def Circuit.withinDegree (circuit : Circuit p) (b : DegreeBound) : Prop :=
...
-  (∀ originalCircuit : Circuit p,
-    let (optimizedCircuit, derivations) := optimizer originalCircuit
-    (optimizedCircuit.isSoundReplacementOf originalCircuit busSemantics) ∧
-    (optimizedCircuit.isCompleteReplacementOf originalCircuit busSemantics
-      derivations))
+  (∀ inputCircuit : InputCircuit p,
+    let (optimizedCircuit, derivations) := optimizer inputCircuit
+    (optimizedCircuit.isSoundReplacementOf inputCircuit.toCircuit busSemantics) ∧
+    (optimizedCircuit.isCompleteReplacementOf inputCircuit.toCircuit
+      busSemantics derivations))
```
georgwiese: I think we should pull this in more, and pass the original circuit as `InputCircuit` into the soundness and completeness statements!

E.g., one consequence of this is that `Derivations.cover` is defined over different variable types (rn both `inputVars` and `outputVars` have the same type), which makes sense.

docs/Docs.lean
```diff
@@ -56,19 +56,23 @@ In the remainder of this document, we formalize the properties that an optimizer
 
 # Variables, expressions and assignments
 
-A {deftech}_variable_ is how the _runtime witness data_ is referenced in a circuit. Variables in the input circuit carry a _powdr ID_, while newly introduced variables do not.
+A {deftech}_variable_ is how the _runtime witness data_ is referenced in a circuit. Every variable of the circuit exported by powdr carries a _powdr ID_, so its type records the ID unconditionally:
```
georgwiese:
```suggestion
A {deftech}_variable_ is how the _runtime witness data_ is referenced in a circuit. Every variable of the circuit exported by powdr carries a _powdr ID_:
```
AI describing diffs

georgwiese:
The −25…−36% runtime win isn't the type change — it's the deep copy of the input circuit in https://github.com/powdr-labs/apc-optimizer/pull/295/files#diff-fc0aa0dfbd62903c927e6c8d44ea48e3054e7865ab2f714c6d3e7ae6045c0e3cR206

On keccak: main 1750 ms, this branch 1410 ms, **main + a 14 ms deep copy of the circuit inside the timed region 1370 ms**. Instructions retired are unchanged (26.39 → 26.61 G, i.e. the copy adds work) while cycles drop 8.46 → 6.96 G — so it's heap layout / cache locality, not less work. Scales with circuit size, hence openvm-eth ≈ 0% and keccak −36%, and hence the flat per-pass `runtime_bench` table.

Worth a line in the PR body: the win is real for FFI callers too, but incidental, so it silently evaporates if the conversion is ever dropped. The spec change itself reads as equivalent-strength, not a weakening (`InputVariable.toVariable` is a bijection onto the ID-carrying `Variable`s).

georgwiese (commented):
Cool! Some comments. Also need to double check whether the docs still make sense, but maybe it is fine after a round of changes.

ApcOptimizer/Spec.lean
Schaeff: I had exactly this but it didn't seem worth it to use `Output` everywhere in the codebase. I don't have a super strong opinion though.

ApcOptimizer/Spec.lean
Schaeff: Agreed

ApcOptimizer/Spec.lean
Schaeff: Right, I wasn't sure how far to pull it in

ApcOptimizer/Implementation/Variable.lean
Schaeff: https://github.com/powdr-labs/powdr/pull/3818

ApcOptimizer/Spec.lean
Schaeff: The PR grew a bit as a result but it does look better

.github/workflows/effectiveness_bench.yml
```diff
@@ -85,5 +85,5 @@ jobs:
             Effectiveness bench of
             ${{ env.BASE_SHA && format('target `{0}` vs baseline `{1}`', env.BENCH_SHA, env.BASE_SHA)
                  || format('`{0}`', env.BENCH_SHA) }}.
-            Circuit sizes are deterministic; any per-case change is real.
+            OutputCircuit sizes are deterministic; any per-case change is real.
```
Schaeff: revert

.github/workflows/lean_action_ci.yml
```diff
@@ -105,7 +105,7 @@ jobs:
   # sha256 stress sets. One cell per benchmark group, and each cell benches **both sides itself** --
   # this PR's binary then the `baseline` job's main build, back to back on the same runner.
   #
-  # Circuit sizes are deterministic, so for effectiveness the sides needn't share a runner. But
+  # OutputCircuit sizes are deterministic, so for effectiveness the sides needn't share a runner. But
```
Schaeff: revert

ApcOptimizer/Spec.lean
```diff
@@ -66,15 +89,15 @@ inductive ComputationMethod (p : ℕ) where
...
   /-- Conditional computation: if `cond` evaluates to zero, use `thenM`, else
       use `elseM`. -/
-  | ifEqZero (cond : Expression p) (thenM elseM : ComputationMethod p)
+  | ifEqZero (cond : InputExpression p) (thenM elseM : ComputationMethod p)
 
 /-- Evaluate a computation method under an assignment (cf. powdr's
     `evaluate_computation_method`). -/
 def ComputationMethod.eval :
-    ComputationMethod p → (Variable → ZMod p) → ZMod p
+    ComputationMethod p → (InputVariable → ZMod p) → ZMod p
```
Schaeff, line 99: we only compute based on input variables

ApcOptimizer/Spec.lean
```diff
@@ -227,22 +277,22 @@ theorem Derivations.methodFor_isSome (ds : Derivations p)
...
   match hp : v.powdrId? with
   -- Well-defined: by the `some` branch of `Derivations.cover`, a powdr-ID
   -- variable of the output circuit also exists in the input circuit.
-  | some _ => inputAssignment v
+  | some id => inputAssignment ⟨v.name, id⟩
   | none => ((ds.methodFor v).get (ds.methodFor_isSome h hv hp)).eval inputAssignment
 -- ANCHOR_END: witgen
 
---------- Circuit implications ---------
+--------- OutputCircuit implications ---------
```
Schaeff: revert

Benchmarks/benchmark.py
```diff
@@ -262,7 +262,7 @@ def emit_compare_md(base, target):
     if not changed:
         lines.append("Per-case circuit sizes are **identical** between main and this branch.")
     else:
-        lines.append(f"Circuit sizes changed on {len(changed)} of {len(common)} cases.")
+        lines.append(f"OutputCircuit sizes changed on {len(changed)} of {len(common)} cases.")
```
Schaeff: revert

ApcOptimizer/Spec.lean
```diff
@@ -7,33 +7,55 @@ variable {p : ℕ} [Fact p.Prime]
 
 --------- Expressions ---------
 
-/-- A circuit variable. -/
-structure Variable where
+/-- A variable in the input circuit. Unlike a `OutputVariable`, it always
+    carries a powdr ID: the input circuit has no derived variables. -/
```
georgwiese, line 11:
```suggestion
/-- A variable in the input circuit. -/
```
Seems noisy

ApcOptimizer/Spec.lean
```diff
@@ -84,15 +106,15 @@ def ComputationMethod.eval :
...
+def ComputationMethod.vars : ComputationMethod p → List InputVariable
   | .const _ => []
   | .quotientOrZero num den => num.vars ++ den.vars
   | .ifEqZero cond thenM elseM => cond.vars ++ thenM.vars ++ elseM.vars
 
 -- ANCHOR: derivations
 /-- A list of derived variables paired with how to compute each, consumed by
     witness generation. -/
-abbrev Derivations (p : ℕ) := List (Variable × ComputationMethod p)
+abbrev Derivations (p : ℕ) := List (OutputVariable × ComputationMethod p)
```
georgwiese, line 117: Nice! That's much clearer than before, very obvious that derivations only exist for output variables and only depend on input variables.

ApcOptimizer/Spec.lean
```diff
@@ -157,28 +179,55 @@ abbrev BusMessage (p : ℕ) := Nat × List (ZMod p)
 abbrev BusState (p : ℕ) := BusMessage p → ZMod p
 -- ANCHOR_END: busState
 
---------- Circuit ---------
+--------- OutputCircuit ---------
```
georgwiese, line 182:
```suggestion
--------- Circuit ---------
```

I think it was fine like this?

ApcOptimizer/Spec.lean
```diff
@@ -338,23 +385,23 @@ def Circuit.withinDegree (circuit : Circuit p) (b : DegreeBound) : Prop :=
...
 
 --------- Optimizer correctness ---------
 
 -- ANCHOR: isCorrect
-/-- An optimizer is correct if, for every input circuit, replacing it with the
-    optimized circuit is both sound and complete, and the optimizer respects
-    the degree bound `b`. -/
+/-- An optimizer is correct if, for every input circuit,
+    replacing it with the optimized circuit is both sound and complete, and the
+    optimizer respects the degree bound `b`. -/
```
georgwiese, line 398:
```suggestion
/-- An optimizer is correct if, for every input circuit, replacing it with the
    optimized circuit is both sound and complete, and the optimizer respects
    the degree bound `b`. -/
```

I think the layout was better before :D

ApcOptimizer/Spec.lean
```diff
@@ -157,28 +179,55 @@ abbrev BusMessage (p : ℕ) := Nat × List (ZMod p)
...
+abbrev OutputCircuit (p : ℕ) := Circuit OutputVariable p
 
 /-- The variables occurring anywhere in a circuit. -/
-def Circuit.vars (circuit : Circuit p) : List Variable :=
+def Circuit.vars {V : Type} (circuit : Circuit V p) : List V :=
   circuit.algebraicConstraints.flatMap Expression.vars ++
     circuit.busInteractions.flatMap
       (fun bi => bi.multiplicity.vars ++ bi.payload.flatMap Expression.vars)
 
+--------- Circuits exported by powdr ---------
```
georgwiese, line 213: As far as I can see, these functions are not actually used in this file, which means it can be moved out of the spec! And with it a bunch of other functions, like `mapVar`! 🎉

I think `Utils/CircuitConversion.lean` would be a good place. It would still be in the audited surface, but in this file (which is the core of the spec), it's unnecessary and a bit distracting.

georgwiese (commented):
Nice! I have a few minor comments, and I think we can move some stuff out of `Spec.lean`!

ApcOptimizer/Spec.lean
georgwiese: Or, does it need to be in the audited surface? Actually not I think! An optimizer is `InputCircuit p → OutputCircuit p × Derivations p`. It can use whatever conversion it wants as long as it can prove correctness.

georgwiese:
Pushed 1ad74c91dda15d906c13254994618af09f44df89 to polish the docs. Not many changes, but I did remove the circuit conversion part already, because I don't think it needs to be part of the spec, as mentioned above.

PR #303, opened 2026-08-13
SP1 under the order-free rely: single-epoch TS_BOUND on clk_low + ENTRY_KEY by pc low limb

1arie1:
Stacked on #288 (same commits up to `b2d3e24`; only the last four commits are new). Opened as a separate PR so CI benchmarks the SP1 support on its own.

This closes #288's declared-open item — "SP1, which declares no timestamp slot". SP1's clock is the two-limb pair `(clk_high, clk_low)` split at bit 24, so no single bounded timestamp slot exists globally. Within one APC block, however, `clk_high` is a *single shared expression* (instruction chips pass it through untouched, and a block needing a `StateBumpChip` carry is never formed as an APC), so in-block order is carried by `clk_low` alone. That makes a *local* TS_BOUND sound:

- **`memTsFieldOf` (audited, `Sp1Semantics.lean`)**: TS_BOUND on slot 1 — `2^24` on memory (true limbs of real access timestamps), `2^25` on the execution bridge (the block's final *sent* state may be the pre-bump carry state, over by less than one instruction's clock increment, which SP1's own `is_clk` booleanity argument caps at `2^24`).
- **`memEntryKeyOf` (audited)**: ENTRY_KEY on the exec bridge by the entry pc's **low limb** — SP1 carries the pc as limbs `[pc mod 2^16, pc / 2^16, …]` in payload slots 2–4, so the designation is slot 2 with `entryPc mod 2^16`; weaker than the full pc but enough to separate the entry record from interior records stepping by 4. Threaded through `sp1Facts`/`sp1Optimizer`/FFI/CLI.
- **Gadget certificates learn two SP1 range-check shapes** (`BusUnify.lean`, shared by `busSweep`): limb bounds through the generic `facts.slotBound` (SP1 checks are 4-slot byte-bus messages, which the two-slot `varRangeBus` scan misses), and one synthetic *expression* limb (`denseBUGadgetX`) — powdr eliminates SP1's `diff_high` witness column and byte-checks its solved expression `(send_ts − recv_ts − 1 − diff_low)·2⁻¹⁶` inside the U8Range payload, so the LessThan certificate subtracts `k·LX` (`k` from the shared variable's coefficient ratio) and certifies the small remainder.
- **`DenseBUIdx`**: a per-invocation candidate-position index (variable ↦ bound-witness positions; variable ↦ `(position, slot)` expression-limb candidates). Untrusted by construction — every consulted entry is re-verified on the single indexed interaction, so a wrong or missing entry costs a miss, never soundness. This replaced the per-variable whole-list scans (first cut of the expression-limb fallback took sp1:keccak past 600 s; with the index it is 2.7 s, effectiveness identical).

The three layers were measured cumulatively: TS_BOUND alone moved nothing (gadget certs died on the byte-bus shapes); certs alone still nothing (without the exec-bridge chain each instruction keeps its own clk base, so no address group shares a ts base); ENTRY_KEY → `execChain` → gauss unifies the bases → the full cascade lands. The same history as OpenVM's on #288: the value is cross-instruction forwarding, and that needs the chain.

Local results (`Benchmarks/benchmark.py`):

| suite | #288 before | this PR | main | powdr |
|---|---|---|---|---|
| SP1 rsp (variables) | 1.838x | **3.868x** | 3.922x | 3.980x |
| SP1 keccak (variables) | 1.704x | **5.042x** | 5.163x | 4.809x |
| OpenVM keccak (variables) | 13.618x | 13.618x | 13.618x | 13.618x |

`Scripts/check-proof-integrity.sh` green (three standard axioms, no unused theorems); build warning-free.

Known residue: rsp trails main by 0.054x (46/100 cases lose small); sp1:keccak constraints 38.4x vs main's 41.9x; sp1:keccak runtime 2.7 s vs main's 0.7 s (joins #288's open busUnify runtime item). Not yet rebased over #301 (busForward), which touches the same subsystem.


1arie1:
Rebased over `main` (`3a66436`, i.e. through #301/#297/#298/#293/#296/#299).

Since the rebase base now contains `busForward` (#301), whose chain lemma takes the *positional* discipline from `facts.admissible_sound`, the pass could not survive the branch's switch to the order-free rely as written. It is therefore **ported rather than merged**: the migration commit temporarily removes it, and the last two commits reinstate it running on `denseBSCanon` — the bus's interactions rearranged into the *certified* canonical access order — with `denseBFCheckCert_sound` taking `admissibleMemoryBus` as a hypothesis supplied by `denseBSOrder?_admissibleMemoryBus`, exactly as `busSweep` already does. `memFwdChain`, the interleaving countermodel and the verifier are otherwise unchanged from #301.

A/B on this branch (with vs without the `busForward` entry), all 303 cases, no regression on any axis:

| suite | variables | bus interactions |
|---|---|---|
| OpenVM keccak | 13.618× → **14.981×** | 7.587× → **7.838×** |
| OpenVM sha256 | 14.191× → **14.430×** | 8.734× → **8.793×** |
| openvm-eth | 4.553× → **4.557×** (wins 31 → 33) | 3.557× unchanged |
| wasm-eth | 7.254× → **7.259×** | 6.205× unchanged |
| sp1/rsp | 3.868× unchanged | 2.692× unchanged |
| sp1/keccak | 5.042× unchanged | 2.987× unchanged |

The four OpenVM aggregates land on `main`'s own post-#301 figures, so the canonical-order detour costs the pass nothing where it fires. Two known items, both recorded in log entry 185: `busForward` and `busSweep` each certify the same canonical order per bus per cleanup cycle (keccak 1.67 → 2.00 s, sha256 22.8 → 30.5 s — the duplicated certification, not the sweep), and the pass has no SP1 effect where #301 measured a small one on `main`.

`Scripts/check-proof-integrity.sh` green; `lake build` warning-free.

georgwiese:
@1arie1 I agree with the spirit of the spec change and would like to do some small iterations on the spec and implementation. Could you change the target branch to `order-free-admissibility` instead of `main`? I also invited you to be a collaborator to the main repo, so you'd still be able to push to the new branch.

georgwiese:
Actually, I think this is an exact duplicate of #288 (points to the same commit), so I'm closing this PR.

1arie1:
I see that you are working on something. I'm not sure whether you still want me to move the PR. Happy to help.

